PR #27, opened 2021-01-07
Header and basic layout

iAdramelk:
Had to replace `@emotion` with `styled-components`, because rehydration problems (`@emotion` was extracting styles to head on server, but  was inlining on client which resulted in different structure of rendered DOM and DOM sent from the server). As a result I needed to replace `rebass` with my own set of base components.

Header is working for desktop. Didn't port mobile styles for now, because original `@media` query logic is differs from the one used in the `styled-system` (`min-width`, vs `max-width`).

Added basic layout.

Styled TOC, but still need to add sticking logic.


components/AnchorNavigation/AnchorNavigation.tsx
```diff
@@ -0,0 +1,54 @@
...
+          <Box key={id} pt={2}>
+            <Link href={`#${id}`}>{title}</Link>
+          </Box>
+        );
+      })}
+    </Box>
+  );
+};
+
+const Link = styled("a")(
```
alex-kovoy, line 39: Please add missing props to the Link component and use that instead
```ts
import { Link } from 'components';

<Link fontSize={12} lineHeight={16} color="xxx" href={`#${id}`}>
  {title}
</Link>
```

components/Navigation/Navigation.tsx
```diff
@@ -16,20 +17,30 @@ interface NavigationProps {
...
-    </ul>
+    <Box
+      width="240px"
+      height="100%"
+      borderRight="1px solid"
+      borderColor="lightest-gray"
+      overflow="auto"
+    >
+      <ul>
+        {data.map(({ title, entries }) => (
```
alex-kovoy, line 28: No need for extra `<ul>` tag because you can use `as` to override an html tag.
```tsx
<Box as="ul"
```

components/Navigation/Navigation.tsx
```diff
@@ -16,20 +17,30 @@ interface NavigationProps {
...
+          <li key={title}>
+            <div>{title}</div>
```
alex-kovoy, line 30: no need for this extra `div` as well

pages/teleport/docs/[[...slug]].tsx
```diff
@@ -29,34 +31,25 @@ const DocsPage = ({
...
-              })}
-            </td>
-          </tr>
-        </tbody>
-      </table>
-    </>
+      <Flex alignItems="stretch">
+        <Box flexShrink={0}>
+          <Navigation data={navigation} />
+        </Box>
```
Comment (alex-kovoy) [inline pages/teleport/docs/[[...slug]].tsx:39]: Let the props pass so you can nicely set the values without adding unnecessary wrappers.

```ts
<Navigation flexShrink={0} data={navigation} />
```

pages/teleport/docs/[[...slug]].tsx
```diff
@@ -1,14 +1,16 @@
 import { GetStaticProps, GetStaticPaths } from "next";
 import { useRouter } from "next/router";
-
+import { Flex, Box } from "components/Blocks";
```
Comment (alex-kovoy) [inline pages/teleport/docs/[[...slug]].tsx:3]:
```tsx
import { 
  Flex,
  Box,
  AnchorNavigation,
  Head,
  Layout,
  Link,
  Navigation,  
  Admonition} from "components";
```

components/Blocks/Blocks.tsx
```diff
@@ -0,0 +1,90 @@
...
+  border,
+  color,
+  flexbox,
+  layout,
+  position,
+  space,
+  typography
+);
+
+export const Box = styled("div")<StyledSystemProps>(
```
alex-kovoy, line 54: Lets create a package `components/system/system.ts` and place all helpers|types in there. The `components/system/index.ts` would re-export `design-system` methods and whatever helpers/types/method we have.

Also, instead of having them all in Blocks.ts, I suggest to create separate components

```
components/Box/
components/Flex/
components/Link/
components/index.ts
```

Yes, Flex and Box looks tiny but Link and Button will grow in functionality.

alex-kovoy (commented):
I see that you started working on the styles which is great! Currently there is a lot of duplication in regards to hyperlinks and typography. To fix it, we can define variants in our scheme object. Please ping me in slack to discuss it (we can do this refactoring in the next PR).

components/Blocks/Blocks.tsx
iAdramelk: Let's place all styled-system stuff in the same folder `system` in root directory.

alex-kovoy (approved):

PR #28, opened 2021-01-08
Docs wrapper

iAdramelk:
- DocsHead component.
- DocsNavigation component.
- Icons in config.
- Version switcher.


iAdramelk:
@alex-kovoy sorry, created PR from master instead of previous branch.

components/DocNavigation/DocNavigation.tsx
```diff
@@ -0,0 +1,147 @@
...
+const DocNavigationCategory = ({
+  id,
+  opened,
+  onToggleOpened,
+  icon,
+  title,
+  entries,
+}: DocNavigationCategoryProps) => {
+  const router = useRouter();
+  const toggleOpened = useCallback(() => onToggleOpened(opened ? null : id), [
```
alex-kovoy, line 36: this optimization `useCallback` is unnecessary. Also consider moving the Category to a separate file (

```
/DocNavigation/
   DocNavigation.tsx
   DocNavigationCategory.tsx
   index.ts
```

components/Icon/icons.ts
```diff
@@ -0,0 +1,17 @@
+export { default as bolt } from "./svg/bolt.svg";
```
alex-kovoy, line 1: lets move it under `svg` folder and rename it to `index.ts`

`index.ts` is a way to access package elements from javascript. The fact that these elements can represent an `svg` file is an implementation detail handled by webpack. So it's ok to use the same index-accessing in this case as well.

```ts
import { book, code, flag } from './svg';
```

alex-kovoy (approved):
with some minor comments

PR #29, opened 2021-01-12
Add styles to markdown

iAdramelk:


components/MDX/Admonition.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const types = ["warning", "tip", "note", "danger"];
+interface AdmonitionProps {
+  type: "warning" | "tip" | "note" | "danger";
+  title?: string;
+  children: React.ReactNode;
+}
+
+const Admonition = ({ type, title, children }: AdmonitionProps) => {
+  const color = types.includes(type) ? type : "tip";
```
alex-kovoy, line 13: instead of `types.includes(type) ? type : "tip";`

I think you can make the `type` optional and do

```ts
  const color = type || "tip";
```

alex-kovoy (approved):
with a comment

components/MDX/Admonition.tsx
iAdramelk: Will not work in this case. This is not guard against `undefined`. its a guard against wrong values. Right now some old `.md`-files have some strange non-standard types and I want to replace them with these 4. But now I think that better idea would be to do it in the remark-plugin instead.

PR #30, opened 2021-01-13
Select styles

iAdramelk:


components/DocHeader/DocHeader.tsx
```diff
@@ -48,7 +48,9 @@ const DocHeader = ({
           <Box as="h1" pr={6} color="white" fontWeight="regular">
             {title}
           </Box>
-          <DocHeaderVersions versions={versions} />
+          <Box mt={1}>
+            <DocHeaderVersions versions={versions} />
```
alex-kovoy, line 52: Lets rename DocHeaderVersions to:

```ts
import Versions, { VersionItem} from "./Versions"
```

This way it matches the file name and also shorter

components/DocHeader/DocHeader.tsx
```diff
@@ -48,7 +48,9 @@ const DocHeader = ({
           <Box as="h1" pr={6} color="white" fontWeight="regular">
             {title}
           </Box>
-          <DocHeaderVersions versions={versions} />
+          <Box mt={1}>
```
alex-kovoy, line 51: You do not need to use a wrapper (div) just to add margin|padding. Pass the props down to component to pick it up.

```ts
 <DocHeaderVersions mt={1} versions={versions} />
 ```

alex-kovoy, line 51: `DocHeaderProps` -> `Props`  No need for this extra prefixes `DocHeaderXXX`, they add no value.

alex-kovoy (approved):
Looks good (with minor comments)

components/DocHeader/DocHeader.tsx
iAdramelk: Have to argue with it. In the editor and error messages in you see a lot of Props that are extending or combining other  Props it's pretty hard to debug them. But if you see the name, it[s pretty easy to find the place of the problem.

PR #32, opened 2021-01-13
Add analytics

iAdramelk:


pages/_app.tsx
```diff
@@ -1,12 +1,39 @@
...
+      Fathom.trackPageview();
+    }
+    // Record a pageview when route changes
+    router.events.on("routeChangeComplete", onRouteChangeComplete);
+
+    // Unassign event listener
+    return () => {
+      router.events.off("routeChangeComplete", onRouteChangeComplete);
+    };
+  }, []);
```
alex-kovoy, line 30: Instead of leaving a comment, it could be expressed as: 

```ts
function cleanup() {
   router.events.off("routeChangeComplete", onRouteChangeComplete);
}

return cleanup;
```

alex-kovoy (approved):

PR #112, opened 2021-02-24
Add simple sitemap generator

iAdramelk:
Based on https://www.sitemaps.org/protocol.html spec.

Will be hosted at the path `/teleport/docs/sitemap.xml` for now.

Some gotchas:

- It includes all existing pages, including previous docs versions, but old versions has lesser priority and update frequency.
- All pages has exact same `lastmod` value – date of the latest build. Thought about using last modified date from mdx file here, but it would prevent it from detecting new plugins and build result updates.


klizhentas (approved):

klizhentas (changes requested):
@iAdramelk there is a lot of duplicate content in the `ver` versions, let's only include the unversioned URLs in the site map

iAdramelk:
> @iAdramelk there is a lot of duplicate content in the `ver` versions, let's only include the unversioned URLs in the site map

Done, removed everything except latest version. Merging now.

PR #164, opened 2021-04-02
Use project root instead of server root.

wadells:
If the docs are not hosted at the server root (e.g. https://goteleport.com/docs/) then we don't want to link to resources above our level.  In this particular case, `/` was linking to the homepage (https://goteleport.com/) instead of the latest docs (https://goteleport.com/docs/).


wadells:
+ @ulysseskan as you recommended the wording updates in 96af0cd.

ulysseskan (commented):
Looks good to me.

iAdramelk:
@wadells thanks for finding this bug. But `"{root}"` will not work in this case. We need to directly set it to `"/docs/"`. Can you please change it?

wadells:
> @wadells thanks for finding this bug. But `"{root}"` will not work in this case. We need to directly set it to `"/docs/"`. Can you please change it?

Won't docs break the previews that aren't behind the nginx reverse proxy?  Admittedly, those aren't nearly as important.

iAdramelk:
@wadells NVM. It's better to fix broken production links asap, so I fixed it myself. Will merge it if staging deploy will finish without problems.

Thanks again.

iAdramelk:
> Won't docs break the previews that aren't behind the nginx reverse proxy? Admittedly, those aren't nearly as important.

@wadells it's ok. We didn't redirect subfolder to the Vercel. Vercel also have `/docs/` folder. It's done to make future porting of the other pages easier

iAdramelk:
It works https://teleport-docs-next-git-walt-latest-link-fix-gravitational.vercel.app/docs/ , merging.

klizhentas:
Screen Shot 2021-04-02 at 2 06 38 PM
Screen Shot 2021-04-02 at 2 06 44 PM

@iAdramelk @wadells needs a couple of spaces after `:` and after `latest` see above

PR #170, opened 2021-04-08
Logical pixels support for images

iAdramelk:
This PR adds a way to mark images as retina-ready.

**Problem**

There are two types of pixels the browser works with: Logical pixels and physical pixels. Logical pixels are the pixels you set in the CSS or HTML when you set a font. E. g. `font-size: 24px;`. On non-retina displays, logical and physical pixels are the same – text with `font-size: 24px` will be rendered as a 24px text. But on retina displays, things are different. For example, on macs, when you set `font-size: 24px`, the browser will render twice as many pixels - 48 to make text look more crisp and detailed.

It works well for the text but not as good for the images. The browser will try to show an image with the same logical pixel number on retina display as on non-retina displays by stretching the image to twice as many physical pixels that will result in blurry images.

One case of such stretching that concerns us is saving screenshots. When you take a screenshot on the retina display, it will be saved in logical pixels (so screenshot made on Mac will be 2x larger than screenshot made on an old PC notebook). But then we insert such an image in the HTML page without specifying sizes, and the browser will stretch it twice more to accommodate for the retina displays, resulting in the large blurry images.

**Solution**

We need to set the image size for the retina images. E. g. if we saved 300x300px image, then we need to add width="150" and height="150" to image tag. But we can't do it fully automatically because we didn't know if the image is a screenshot or not or if it was taken on the retina or non-retina display. So we need to add this info to our screenshots.

We can do it by adding a suffix to the images. E. g. `filename@2x.png` means that the screenshot was taken on the retina display with a 2x physical pixel ratio and should be scaled down. In can be other values to accommodate different values: `@3x` will reduce image size by 2/3, `@1.5x` by 1/3, and so on.

This PR adds support for this filename syntax and corresponding tests.


utils/rehype-images.ts
```diff
@@ -6,51 +6,69 @@ import visit from "unist-util-visit-parents";
...
 
         try {
           const { width, height } = probe.sync(file);
+          const scaleRatio = getScaleRatio(src);
 
-          if (node.properties) {
-            node.properties.width = width;
-            node.properties.height = height;
-          }
+          node.properties.width = width / scaleRatio;
```
Saionaro, line 59: What do you think about round the division result to integers? `Math.floor()` for example

iAdramelk: I thought about, but then decided that modern browsers can work with fractional values and browser will probably scale it better with an exact value then with rounded one. Also if it really is a retina image, it should be dividable by this value. If it is not dividable, you probably shouldn't try to divide it in the first place.

Saionaro: Yep, it makes sense 👍

Saionaro (approved):

PR #176, opened 2021-04-17
Add Tile component

iAdramelk:
Added two component:

- `<Tile>` - see https://app.avocode.com/revisions/77798605/?design=51945807.
- `<TileSet>` - to correctly group multiple tiles together.

Desktop version:

image

Mobile version:

image

Can be checked in `storybook`. Run `yarn storybook` for that.


components/Tile/Tile.tsx
```diff
@@ -0,0 +1,109 @@
+import styled from "styled-components";
+import css from "@styled-system/css";
+import Box from "components/Box";
+import Flex from "components/Flex";
+import Icon, { IconName } from "components/Icon";
+import Link from "components/Link";
+
```
Saionaro, line 7:
```suggestion
```

components/Tile/index.ts
```diff
@@ -0,0 +1,3 @@
+export { default, TileSet } from "./Tile";
+
```
Saionaro, line 2:
```suggestion
```

components/Tile/Tile.tsx
```diff
@@ -0,0 +1,109 @@
...
+import bgHref from "./assets/bg.svg";
+
+export interface TileSetProps {
+  children: React.ReactNode;
```
Saionaro, line 11:
```suggestion
  children: ReactElement<typeof Tile> | Array<ReactElement<typeof Tile>>;
```
It's can be useful if we specify a little bit stricter types here.

PR #178, opened 2021-04-19
Add PlanCard component

Saionaro:
plus some icons for future components

Screenshot 2021-04-21 at 18 40 04


utils/string.ts
```diff
@@ -24,3 +24,28 @@ export function camelCaseToDash(str: string): string {
 
 export const toSnakeCase = (str: string): string =>
   str.toLowerCase().replace(/\s/g, "_");
+
+export function formatMoney(value: number | string, divider = ","): string {
```
iAdramelk, line 28: What do you think about using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat ?

components/PlanCard/PlanCard.tsx
```diff
@@ -0,0 +1,170 @@
...
+function renderRecommendedLabel(): ReactNode {
+  return (
+    <Box
+      as="span"
+      text="text-sm"
+      textTransform="uppercase"
+      fontWeight="bold"
+      borderRadius="circle"
+      color="white"
+      backgroundColor="green"
```
iAdramelk, line 104:
```suggestion
      bg="green"
```

components/PlanCard/PlanCard.tsx
```diff
@@ -0,0 +1,170 @@
...
+        variant={accented ? "primary" : "secondary"}
+        shape="lg"
+      >
+        {action.title}
+      </Button>
+    </Flex>
+  );
+}
+
+function renderRecommendedLabel(): ReactNode {
```
iAdramelk, line 95: Why not make `render*` functions components?

utils/string.ts
Saionaro: Sure. But I used to think about `intl` as a set of weak performance methods.
Is something changed in recent years?

components/PlanCard/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  title: string;
+}
+
+export enum PlanIcon {
+  DOWNLOAD = "download",
+  SHIELD = "shieldCheck",
+  BUSINESS = "apartment",
+}
+
+export enum BillingKind {
```
alex-kovoy, line 12:
```ts
export type BillingKind = "monthly" | "free" | "custom";
```

(nit) this would be more preferable (shorter and cleaner)

components/PlanCard/PlanCard.tsx
```diff
@@ -0,0 +1,115 @@
...
+  accented?: boolean;
+}
+
+export function PlanCard({
+  data,
+  accented,
+  charge,
+  hideBadge,
+  ...props
+}: PlanCardProps & FlexProps) {
```
alex-kovoy, line 24:
```ts
export interface PlanCardProps extends FlexProps {
}
```

components/PlanCard/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  MONTHLY = "monthly",
+  FREE = "free",
+  CUSTOM = "custom",
+}
+
+export enum CardHeight {
```
alex-kovoy, line 18: i think its OK to hardcode this values (CardHeight) without a typing them. Unless, of course, if these constants are used in multiple places.

components/PlanCard/types.ts
```diff
@@ -0,0 +1,47 @@
+interface Feature {
+  id: string;
+  title: string;
+}
+
+export enum PlanIcon {
```
alex-kovoy, line 6: Do we have 1 place for all icon names? Because that would be a preferable way of referencing  them.

PR #184, opened 2021-04-19
Checkbox component

Saionaro:


components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,112 @@
...
+const StyledSquare = styled("div")<StyledWrapperProps>(
+  css({
+    display: "flex",
+    alignItems: "center",
+    justifyContent: "center",
+    boxSizing: "border-box",
+    width: 24,
+    height: 24,
+    border: "2px solid",
+    borderColor: "dark-purple",
```
iAdramelk, line 67: Better move to selected element. Or they both will be inside the same css rule in `unselected` case.

components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,112 @@
...
+    backgroundColor: "white",
+    transition: transition([
+      ["borderColor", "interaction"],
+      ["color", "interaction"],
+    ]),
+    borderRadius: "default",
+    "&:focus, &:hover": {
+      borderColor: "light-purple",
+      color: "light-purple",
+    },
```
iAdramelk, line 77: Aren't these redundant if we always have `StyledWrapper`?

components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,112 @@
...
+        width={0}
+        height={0}
+      />
+      <StyledSquare
+        variant={checked ? "selected" : "unselected"}
+        mr={children ? "2" : "0"}
+      >
+        {checked && (
+          <Icon name="check" size="sm" color="dark-purple" lineHeight="sm" />
+        )}
```
iAdramelk, line 45: Nitpick: I would make it with opacity instead of removing element from DOM. This way we can also animate it.

components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+export default {
+  component: Component,
+  title: "Checkbox",
+};
+
+export const Plain = StoryComponent.bind({});
+
+Plain.args = {
+  checked: false,
+};
```
iAdramelk, line 27: Can you please add example with label and add `onClick` to actions https://storybook.js.org/docs/react/essentials/actions#action-args

iAdramelk (changes requested):

PR #188, opened 2021-04-20
Add page intro responsive styles

Saionaro:


components/system.ts
```diff
@@ -51,6 +51,7 @@ export interface StyledSystemProps
   textDecoration?: ResponsiveValue<Property.TextAlign, RequiredTheme>;
   textTransform?: ResponsiveValue<Property.TextTransform, RequiredTheme>;
   listStyle?: ResponsiveValue<Property.ListStyle, RequiredTheme>;
+  borderCollapse?: ResponsiveValue<Property.BorderCollapse, RequiredTheme>;
```
iAdramelk, line 54: Did you miss also adding it to the `system` below or it works without it?

components/Layout/Centrator.tsx
```diff
@@ -1,9 +1,17 @@
...
 export function Centrator({ children, as = "div", ...props }: FlexProps) {
   return (
-    <Flex px="11" justifyContent="center" width="100%">
-      <Flex as={as} maxWidth={1240} {...props}>
+    <Flex
+      px="11"
+      justifyContent="center"
+      width="100%"
+      css={media("md", {
+        px: 3,
```
iAdramelk, line 11: Maybe it's easier to use `[3,3,0]` here?

iAdramelk (commented):
P. S. Can you add screenshots of the components to the PRs? Its easier to check them this way.

components/system.ts
Saionaro: Actually I forget about it! But it works even without adding :|

components/Layout/Centrator.tsx
Saionaro: Wow, it's great

PR #192, opened 2021-04-21
Add email subscription form

Saionaro:
Screenshot 2021-04-21 at 17 13 02


utils/hooks.ts
```diff
@@ -0,0 +1,47 @@
+import { useState, useEffect } from "react";
```
iAdramelk, line 1: I'd either rename this file to `munchkin.ts` or move all munchkin related logic to separate file in utils and only leave hooks-related logic here.

Righto now most of the code not about the hooks, but about muchkin-related logic.

components/EmailSubscribe/EmailSubscribe.tsx
```diff
@@ -1,7 +1,23 @@
...
+import styled from "styled-components";
+import css from "@styled-system/css";
+import { transition } from "components/system";
+import { useMarketo } from "utils/hooks";
+
+const MARKETO_DATA = {
+  baseUrl: "//app-ab44.marketo.com",
+  munchkinId: "819-WHT-483",
+  formId: "1027",
+};
```
iAdramelk, line 12: how about moving it to `.env` files? Also `baseUrl` and `munchkinId` probably not change between different forms, so we can add them inside utils as default values instead of sending them each time in component.

components/EmailSubscribe/EmailSubscribe.tsx
```diff
@@ -14,10 +30,66 @@ export function EmailSubscribe({ ...props }: FlexProps) {
...
-      <Box ml={[0, 2]} mt={[2, 0]}>
-        [email form placeholder]
+      <Box
+        ml={[0, 4]}
+        mt={[4, 0]}
+        minHeight="40px"
+        minWidth="340px"
+        bg={formLoaded ? "transparent" : "lightest-gray"}
+        css={css({
+          transition: transition([["backgroundColor", "interaction"]]),
```
iAdramelk, line 40: Do you need `css` here? It should work with just `transition` prop.

components/EmailSubscribe/EmailSubscribe.tsx
Saionaro: Yeach, great note

iAdramelk (approved):

PR #193, opened 2021-04-21
Add TileImage component

iAdramelk:
Usage example:

```
<TileImage
  title="SSH & Nodes"
  src="../img/azuread/azure-11-loginwithmsft.png"
  href="https://goteleport.com/"
>
  Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris sed dui
  consequat, sagittis.
</TileImage>
```

Render:

image


klizhentas (approved):

components/Tile/TileImage.tsx
```diff
@@ -0,0 +1,113 @@
...
+        <Link
+          href={href}
+          css={css({
+            transition: transition([["opacity", "interaction"]]),
+            "&:hover, &:active, &:focus": { opacity: "0.9" },
+          })}
+        >
+          <Image
+            src={src}
+            alt=""
```
Saionaro, line 42: WDYT to pass something like a `Image ${title}` prop here? Or even add a special `alt` prop?

iAdramelk: I thought about it, but we have title in the plain text on the next line so it would be redundant (and screen reader just read it twice in the row). Also looking at the current screenshots, they mostly work as illustration for the title below. So again adding alt for them would also be redundant in the most cases. Can't though of meaningful alt texts for them, tbh.

Saionaro: For example, for code screenshots we can pass plaintext code as a alt to image. It will be VERY useful.

iAdramelk: Lets add `alt` as an optional field then.

PR #197, opened 2021-04-22
Add storybook deploy

iAdramelk:
Should be available at `/storybook/`.


klizhentas:
@iAdramelk let's make sure `/storybook` is only available on dev view (and linked on the website), not a prod version.

iAdramelk:
> let's make sure /storybook is only available on dev view (and linked here), not a prod version.

@klizhentas can you give me more details? Right now it deploys with the docs site in each branch as `/storybook/`. It will not be visible on the main `goteleport.com` site if you don't add it to redirects manually. Should I remove it from the `main` branch deploy?

iAdramelk:
@klizhentas updated to only deploy for previews.

.eslintignore
```diff
@@ -2,4 +2,5 @@
 /.next
 /node_modules
 /pages/ver/*
-/content
\ No newline at end of file
+/content
+/public/storybook
```
Saionaro, line 6:
```suggestion
/public/storybook

```

.prettierignore
```diff
@@ -2,4 +2,5 @@
 /.build
 /node_modules
 /pages/ver/*
-/content
\ No newline at end of file
+/content
+/public/storybook
```
Saionaro, line 6:
```suggestion
/public/storybook

```

PR #198, opened 2021-04-22
Plans calculator page

Saionaro:


Saionaro:
@iAdramelk sorry, this one a little big, but it's atomic.

It's not a final version though. But I'd like if you started some review when you have time.

klizhentas (changes requested):
@Saionaro I will let @alex-kovoy and @iAdramelk to review the code and @alexwolfe to comment on the design. My review is regarding end user experience.

Users of the webpage will be interacting with the contents of 

pages/pricing/index.tsx

to tweak the title, typos and parts of the component.

What they would expect is a page similar to this:

```markdown
---
title: Teleport Adopters
description: Who is using Teleport?
h1: Teleport Adopters and Use-Cases
---

## Who is using Teleport?

Whether you are using Teleport OSS, Enteprise or Cloud, tell us about your use-case.
Just click "Improve the docs" and send us a PR.

<Head title="Pricing" />

      <Layout>
        <Centrator wrapperAs="section">
          <PageIntro data={INTRO_DATA} />
        </Centrator>
        <PlanCalculator as="section" />
      </Layout>

      <Footer />
```

In this mode you would still use the react component you created but to the end users and content editor it will be presented as mdx format file, not a tsx file that is a bit scary for non-typescripters to read.

alex-kovoy (changes requested):
Lets stop using dynamic IDs and remove everything that is not required. Please follow this sample to clean up the rest https://github.com/gravitational/next/pull/199

I suggest to start refactoring by deleting the `PlanCalculator/utils.ts` file.

A few other notes:

I see an overuse of `const XXX` statement where a lot of unnecessary things are defined as CONST. Please try to inline most of it. Also, do not use it for objects/json.

Saionaro:
@klizhentas got it. Would you like if plans/footer links/options/etc descriptions will be available in md as well?

One more thing: could you please add me to [js-style-guide](https://github.com/gravitational/js-style-guide) repo? I guess it's important to me to learn your code-style for an interaction improvement :)

@alex-kovoy ok, let's use indexes instead of ids if it's more reasonable to you.

klizhentas:
> @klizhentas got it. Would you like if plans/footer links/options/etc descriptions will be available in md as well?

Yes, it will enable our product marketing team to make tweaks with no coding

> One more thing: could you please add me to js-style-guide repo? I guess it's important to me to learn your code-style for an interaction improvement :)

Done

Saionaro:
@alex-kovoy here we go: https://github.com/gravitational/next/pull/202
Variant still requires a serial id for fash access to avoid a tightrope-walking while price calculating.
https://github.com/gravitational/next/blob/55d07243235a8ae741121a520dac89fa5eed7e5e/components/PlanCalculator/structure.ts#L112-L120

alex-kovoy:
> Variant still requires a serial id for fast access to avoid a tightrope-walking while price calculating.
> https://github.com/gravitational/next/blob/55d07243235a8ae741121a520dac89fa5eed7e5e/components/PlanCalculator/structure.ts#L112-L120

Thanks, I will take a look. Just for the records, I am not against `ids` as a concept, I am against _code_ in general. The less there is the better :)

Saionaro:
@alex-kovoy Okay, I understood your vision.
The only note: it's possible to have performance issues while slider moving. Especially on low end devices.

alex-kovoy:
Premature optimization is a root of all evils. Not only we do not know the actual performance numbers, but also we are not sure if there ever be a user with a slow device trying to use enterprise calculator. Lets be a bit more practical. I would even go further and recommend you to stop even worrying about performance. For now.

By the way, I am pretty sure there wont be any significant/noticeable performance differences between 2 solutions due to such a small data size.

Saionaro:
Even I on my MBP2016 noticed a significant performance drop in new version:
https://www.youtube.com/watch?v=WmITRLDnnaQ
The previous version I recorded here:
https://www.youtube.com/watch?v=KGL942pW1Kg

If we say performance is not a key here, let it be.

alex-kovoy (approved):

alex-kovoy:
@Saionaro  Oh, this is because of the Slider component where I forgot about throttling behavior that is disabled now.  Could you fix it by adding a `RowAccess`? 

Or you can just add this as a quick fix which will bring the performance back (tested on the 11 laptop) :) 

```tsx
  const notify = useMemo(() => {
    return throttle(200, false, (price = 0) => {
      onChange(price);
    });
  }, []);

  useEffect(() => {
    notify(calc.price);
  }, [calc.price]);`
```

> If we say performance is not a key here, let it be.

Please not ignore _the basic throttling_ behavior.

alex-kovoy:
And just to clarify, choosing between premature optimization, performed on the low quality components VS not doing optimization _at all_, lets do the later.

PR #203, opened 2021-04-23
Update remark-links to work with JSX elements

### LINKED ISSUE #200: Tile component should render the URL
I have tried:

```
<TileSet>
<Tile href="./getting-started.mdx" icon="bolt" title="Get started with Open Source">
  Install Teleport Open Source [in 5 minutes](getting-started.mdx).
</Tile>
</TileSet>
````

Next generated the URL: `/docs/ver/6.2/getting-started.mdx`

I expected it to process the mdx file, verify that it exists and generate the link:

`/docs/ver/6.2/getting-started`

### PR DESCRIPTION
**Pull request by iAdramelk**
Fix #200: Add support for `href` prop rewrite in components.

See https://github.com/gravitational/next/compare/fix/component-href-values?expand=1#diff-ac21dd23114cde8ebc5361e92753911f8b9b5feb9a3634a0b6d0655d6097937c for usage examples.

To make it work had to remade it as remark plugin instead of rehype one. And update type definitions.

### CODE REVIEW

**Review by klizhentas — APPROVED** [review]

**Comment by klizhentas** [comment]
@Saionaro can we please review and merge on Monday, this will unlock me with tiles.

File: utils/remark-links.ts [hunk 1]
```diff
@@ -0,0 +1,66 @@
...
+
+    visit<Link | MdxAnyElement>(root, [isLocalLink], (node) => {
+      if (node.type === "link") {
+        node.url = updateHref(basename, node.url);
+      } else if (mdxNodeTypes.includes(node.type)) {
+        const hrefAttribute = node.attributes.find(
+          ({ name }) => name === "href"
+        );
+
+        hrefAttribute.value = updateHref(
```
Saionaro, line 59: It looks like it's possible to have `hrefAttribute === null` here. Shouldn't we add check here?

iAdramelk: Actually ,it should always exists: Second argument of `visit` is filtering function. So we only enter nodes that matches `isLocalLink`. ANd it's either mdast link node or mdx node with href attribute. 

And we already resolved link nodes in the first part of the if. So only mdx nodes with href attribute should be left.

But maybe if I replace `if` statement with `isMdxComponentWithLocalHref` and `isRemarkLinkWilthLocalHref` it will be easier to read.

iAdramelk: Ok, I actually remembered why I haven't done it. `node.type` check will correctly detect node type, but function will not. So if I change it to `isRemarkLinkWilthLocalHref`, then will need to do manual typecasting. Will leave it as is for now.

utils/create-mdxjsesm-node.ts
```diff
@@ -29,7 +30,7 @@ const createMdxjsEsmNode = (name: string, data: Value) => {
         ],
       },
     },
-  };
+  } as EsmNode;
```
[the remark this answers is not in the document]
iAdramelk: Need typecasting here, I afraid. Without it I will get this error on build:

```
error TS2345: Argument of type '{ type: string; value: string; data: { estree: { type: string; sourceType: string; body: { type: string; source: any; specifiers: any[]; declaration: { type: string; kind: string; declarations: { type: string; id: { ...; }; init: Expression; }[]; }; }[]; }; }; }' is not assignable to parameter of type 'MdxastNode'.
```

But its probably better to move to the top: `const createMdxjsEsmNode = (name: string, data: Value): EsmNode => {`.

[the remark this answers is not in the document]
iAdramelk: Ok, missed first part and feel stupid now.

PR #204, opened 2021-04-23
Fix #201: Wrong padding on standalone Tiles

### LINKED ISSUE #201: Missing padding when combining tiles
When using two tile sets alongisde each other and other components - like embedded videos, there is no padding
Screen Shot 2021-04-22 at 8 19 05 PM

iAdramelk:
Should fix #201. See screenshot:

image

CC @klizhentas If it still persists after merge, please let me know.


klizhentas:
@Saionaro can we please review and merge on Monday, this will unlock me with tiles.

components/Tile/TileSet.tsx
```diff
@@ -1,17 +1,44 @@
...
 }
 
 const TileSet = ({ children }: TileSetProps) => {
   return (
-    <Flex flexWrap="wrap" alignItems="stretch">
+    <Flex
+      flexWrap="wrap"
+      alignItems="stretch"
+      css={css({
+        [`& ${StyledWrapper}:nth-child(3n+2), & ${StyledWrapper}:nth-child(3n)`]: {
```
Saionaro, line 37: We can also consider this one selector:
```suggestion
        [`& ${StyledWrapper}:not(:nth-child(3n+1))`]: {
```
it looks a little bit simpler to me

iAdramelk: We need to also change default margin and margin value inside this selector to make it work.

Right now `3n+1` will select 1, 4, 7, etc elements. But original selector was selecting all elements except these ones.

Will change it in original code.

PR #205, opened 2021-04-24
Clean up

alex-kovoy:
I did a quick round of refactoring+cleanup. 

Here is some notes:

instead of

```ts
import Box from 'components/Box';
import Flex from 'components/Flex';
```

please use
```ts
import { Box, Flex } from 'components';
```

We need to move fast and many of these components will be changed in the future anyways.

Lets stop spending our time on premature optimization. 

Prefer to write simple and dummy code.


alex-kovoy:
@Saionaro please review

components/PlanCalculator/Calculator/useCalculator.ts
```diff
@@ -0,0 +1,87 @@
+import { useState } from "react";
+import getItems, { ItemOption, ItemAccess } from "./getItems";
+
+export default function useCalculator() {
```
Saionaro, line 4: Not sure if this particular file is really simplification against previous calculation logic version

alex-kovoy: @Saionaro Previous one was a collection of IDs->Value maps, which was another abstraction build upon existing objects. Reading these sets of `index->value` maps and use them for actual calculations felt hard to reason about. Event writing a test was not as straightforward as to compare JSON objects.

As for the writing extra code, If you put together all the deleted files into 1 file, with hardcoded constants, perhaps it would better illustrate my point. 

Anyway,  as a general recommendation, please avoid using unnecessary data structures, types, CONSTs, npm packages, etc and try to keep code simple.

Issue #194, opened 2021-04-21
Images that zoom in or enlarge when clicked


inertial-frame:
Potentially related to: https://github.com/gravitational/next/issues/158

Is it possible to add the following?

1. We currently compile MDX files (and their associated static resources) through a next plugin.
2. Can we also inject into the resultant HTML/DOM some kind of JS event or element that allows an image to expand so the image/graph/chart can be inspected at a different viewing scale?

I think this would be helpful since several images are very well-done but are somewhat difficult to read in their current and singular size.

iAdramelk:
Yes, definitely. We can, and should, do it.

PR #206, opened 2021-04-26
Add Figure component

iAdramelk:
Adds `<Figure>` component wrapper for images:

``` 
<Figure align="center" caption="Image caption" bordered>
  ![Alt text](path-to-image.png "Title value")
</Figure>
```

Will render as:

image

See storybook for additional examples.

CC @alexwolfe Are you ok with current styles for border and caption on screenshot or should we change it to something?
CC @inertial-frame Are you ok with this syntax or should we replace it with something?


components/MDX/Image.tsx
```diff
@@ -1,21 +1,77 @@
...
+export const Image = ({
+  align,
+  bordered,
+  caption,
+  ...props
+}: ImageProps & NextImageProps) => {
+  const imageProps = useMemo((): NextImageProps => {
+    return {
+      ...props,
+      width: parseInt(props.width as string, 10),
```
Saionaro, line 33: Shouldn't we specify `width` and `height` as required props?

components/MDX/Image.tsx
```diff
@@ -1,21 +1,77 @@
...
 
-export default Image;
+Image.defaultProps = {
+  align: "left",
+};
+
+type ImageComponent = React.ReactElement<typeof Image>;
+
+export interface FigureProps extends ImageProps {
+  children: ImageComponent[];
```
Saionaro, line 70:
```suggestion
  children: ImageComponent;
```
Instead of `only` in line 74?

components/MDX/Image.tsx
iAdramelk: Weeeell, it's complicated.

`<Image />` shouldn't ever be used manually. It should always be used as an `img` tag replacement inside the `mdx`. And all `img` elements in `mdx` should always have `width` and `height` because we add them automatically by plugin.

So the only two situation then we didn't have them are: 1) Somebody used `<Image>` component directly. 2) Image file is damaged and we can't read its sizes.

But to make component's API more clear I probably add some logic here to address situation then we didn't have width and height.

components/MDX/Image.tsx
iAdramelk: I will commit this one, but we shouldn't remove `only` – it's here to throw errors if we have some additional children inside markdown file.

components/MDX/Image.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+  title: "Docs/Figure",
+};
+
+export const Center = Component.bind({});
+
+Center.args = {
+  align: "center",
+  caption: "Caption example",
+  children: (
+    <Image src={imageSrc} width="581" height="265" alt="" unoptimized />
```
alexwolfe, line 19: Any reason the width and height are hard coded?

iAdramelk: It's a storybook example, so I added sizes here to make it look better (scaled actual sizes to half width/height to make different aligns visible).

In actual component sizes are needed if we want to correctly use different srcset values to show different versions for retina/non-retina/mobile etc. but if we just want to show image as is, then we can omit them. 

P. S. In markdown files these sizes are set automatically at the build step for each local image by reading image headers and adding width/height, so in actual docs page for example, we always try to load best image size for the current screen.

inertial-frame:
Looks great so far! QQ's:

1. Are there plans to incorporate something like the clicking functionality from: https://github.com/gravitational/next/issues/194? I think that'd be very helpful.
2. Do we use the same `width` and `height` values for mobile? If not, it might make sense to move those values to relative ones and use the current one's as defaults. Maybe using JS `|` pipe operator or defining a JS object for CSS-in-JS (or just using a string).
3. I noticed that we have a lot of static assets in the repo, do we ever plan on moving these to a CDN? Is there much benefit to doing so? Does doing so alter any of the image settings?
4. Per the Q above, does it make sense to use something akin to: https://novelist.xyz/tech/progressive-background-image-loading/?  Lazy-loading images (or any rendered static asset for that matter)?

iAdramelk:
@inertial-frame 

> 1. Are there plans to incorporate something like the clicking functionality from: #194? I think that'd be very helpful.

I will add it later, but maybe not to this component, but to original image. Still thinking about the best way to do it.

> 2. Do we use the same width and height values for mobile? If not, it might make sense to move those values to relative ones and use the current one's as defaults. Maybe using JS | pipe operator or defining a JS object for CSS-in-JS (or just using a string).

Yes, width/height here are not used directly. They are supplied to the [next/image](https://nextjs.org/docs/api-reference/next/image) which is then decides what image sizes to create for using on retina and what size to show right now.

> 3. I noticed that we have a lot of static assets in the repo, do we ever plan on moving these to a CDN? Is there much benefit to doing so? Does doing so alter any of the image settings?

We don't have plans ATM as far as I know. We use Vercel for deployments, which works as CDN for the end users (it places all static files to CDN). But I didn't hear about plans for moving images from the repo itself. If we add it someday. I will need to change couple of things in the Nextjs config and inside `remark-images` plugin that I wrote, but this component itself will work as is.

> 4. Per the Q above, does it make sense to use something akin to: https://novelist.xyz/tech/progressive-background-image-loading/? Lazy-loading images (or any rendered static asset for that matter)?

`next/image` uses `loading="lazy"` attribute by default. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading So we only load images that thy are not that far from the current screen position (position value changes from browser to browser, Firefox starts loading them then they are almost on he screen, Chrome loads images  earlier).

We can make some measurements, but for now I think it is more or less enough.

PR #207, opened 2021-04-26
MDX pricing

Saionaro:


pages/pricing/index.mdx
```diff
@@ -0,0 +1,18 @@
...
+import PlanCalculator from "components/PlanCalculator";
+
+<Centrator wrapperAs="section">
+  <PageIntro
+    title={meta.h1}
+    subject={meta.subject}
+    description={meta.description}
+  />
+</Centrator>
+<PlanCalculator as="section" />
```
alex-kovoy, line 18: @iAdramelk 
Why is the content for the website is under `pages/pricing/index.mdx` but the content for the docs are under `content/` ? Just wonder what are the plans for structuring the content.

alex-kovoy: From the content creator, the following UX is preferable. 

```tsx
<Pricing 
    ...
    title={meta.h1}
    subject={meta.subject}
    description={meta.description}
/>
```

pages/pricing/index.mdx
```diff
@@ -0,0 +1,18 @@
...
+h1: Get Started
+subject: Plans and Pricing
+description: Teleport is available either as a free, open source download, or through our commercial offerings which can be hosted by us or self-hosted. The pricing is determined by the number of computing resources accessible via Teleport.
+---
+
+import { PageIntro } from "components/PageIntro";
+import PlanCalculator from "components/PlanCalculator";
+
+<Centrator wrapperAs="section">
+  <PageIntro
```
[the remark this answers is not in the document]
Saionaro: `Centrator` provides two `as` props: for both outer and inner nodes.

pages/pricing/index.mdx
Saionaro: 1. @klizhentas asked for that particular UX. 
2. Using the UX a content creator have more power over page content - add another blocks, change blocks order etc. Not just a fixing some text.

iAdramelk: @alex-kovoy all pages are under `pages` folder actually. But ones for the docs are symlinked here from the `content`. We have content folder because we need a place to init submodules to, but because we only need `mdx` files from here and don't need rest of the folders and files to be available publicly we can't just init them to `pages` folder itself. 

So we init submodules to `content`, then symlink `content/X.X/docs/pages/**/*.mdx` to `pages/docs/XX/**.*mdx`.

But maybe its better to rename `content` folder to something like `submodules` to prevent confusion.

components/MDX/MDX.tsx
```diff
@@ -47,6 +49,9 @@ export const components = {
   TileList,
   TileListItem,
   TileImage,
+  Layout,
+  Centrator,
+  PageIntro,
```
iAdramelk, line 54: I'm afraid of adding too much stuff to provider, because of bundle size increase. Maybe we should move everything except html tags to import. Need to discuss it.

utils/mdx-config.ts
```diff
@@ -1,39 +1,26 @@
...
 const DEFAULT_RENDERER = `
 /** @jsxRuntime classic */
 
-import { mdx } from "@mdx-js/react";
-import DocsPage from "components/DocsPage";
+import { mdx, MDXProvider } from "@mdx-js/react";
+import { Footer, Head, MDX } from "components";
+import { components } from "components/MDX";
+import Layout from "components/Layout";
+import wavePngUrl from "sharedAssets/images/wave-light.png";
```
iAdramelk, line 12: I think we need to create components for different types of layouts (E. g., `SitePage` component here`) and only use resulting components here. The imports and stuff  we have inside the string, the easier it will be break something.

components/MDX/MDX.tsx
iAdramelk: Also, you don't need to add them here if you use `import` in the mdx file.

pages/pricing/index.mdx
```diff
@@ -0,0 +1,17 @@
+---
+title: Pricing
+h1: Get Started
+subject: Plans and Pricing
+description: Teleport is available either as a free, open source download, or through our commercial offerings which can be hosted by us or self-hosted. The pricing is determined by the number of computing resources accessible via Teleport.
+---
+
+import PlanCalculator from "components/PlanCalculator";
```
iAdramelk, line 8: Lets add all components that we want to use in mdx files to `components/index.tsx` and always import them as `import { ComponentName, OtherComponent } from "components"`. This way it will be easier to the people unfamiliar with our components to add them. Right now some components have named exports and some have `default` exports, so without checking component itself it is hard to know how to ad it to the page.

pages/pricing/index.mdx
```diff
@@ -0,0 +1,17 @@
...
+
+<Centrator wrapperAs="section">
+  <PageIntro
+    title={meta.h1}
+    subject={meta.subject}
+    description={meta.description}
```
iAdramelk, line 14: Do we need to use `frontmatter` here? Why not just write props directly?

utils/mdx-config.ts
```diff
@@ -1,39 +1,26 @@
...
-    meta={meta}
-    navigation={navigation}
-    versions={versions}
-    tableOfConents={tableOfConents}
-    githubUrl={githubUrl}
-  >
-    <MDXContent />
-  </DocsPage>
+  <>
+    <Head title={meta.title} />
```
iAdramelk, line 16: We will need to be able to set other meta info for the page. Not only title.

utils/mdx-config.ts
```diff
@@ -47,39 +34,8 @@ interface MdxConfig {
...
-      },
-    ],
-    [
-      rehypeHighlight,
-      { aliases: { bash: ["bsh", "systemd"], yaml: ["conf", "toml"] } },
-    ],
-    [rehypeHeaders, { maxLevel: 2 }],
-  ],
+  remarkPlugins: [remarkFrontmatter, remarkImportFrontmatter],
+  rehypePlugins: [],
```
iAdramelk, line 38: You can remove `rehypePlugins` here, this is not required field.

pages/pricing/index.mdx
Saionaro: @klizhentas WDYT about @iAdramelk 's suggestion? Do you think if current variant more convenient here?
I mean that one:
```
---
title: Pricing
...
---


...
title={meta.h1}
...
```
vs
```
...
title="Pricing"
...
```

alex-kovoy (approved):

PR #223, opened 2021-04-29
add server icon

inertial-frame:
Adds `server` icon for new "Server Access" section. 

---

screen_sm


Saionaro:
Just a little note - it would be great to use `currentColor` keyword instead of just `black`. It will allow us to change an icon color during development.

inertial-frame:
@Saionaro - Ah! Thanks - made the icon fill and added that. Also slightly improved it!

klizhentas (approved):

PR #249, opened 2021-05-19
Add banner component

Saionaro:
Screenshot 2021-05-19 at 15 23 55


components/Banner/Banner.tsx
```diff
@@ -0,0 +1,25 @@
...
+  title?: string;
+  subtitle?: string;
+} & BoxProps;
+
+export default function Banner({ title, subtitle, ...props }: Props) {
+  return (
+    <Box
+      as="section"
+      py={[6, 11]}
+      background={`url(${spacecraftUrl}) center bottom no-repeat`}
```
iAdramelk, line 16: Maybe we want to use `next/image` here to create mobile size version?

iAdramelk (approved):

components/Banner/Banner.tsx
Saionaro: Shall we forced to use it as img tag then? I thought to just add a second static mobile version.

PR #255, opened 2021-05-20
Refactor top menu and add Sign In dropdown

iAdramelk:


components/Header/HeaderCTA.tsx
```diff
@@ -0,0 +1,98 @@
...
+            right={[0, 3]}
+            position={["relative", "absolute"]}
+            width={["100%", "auto"]}
+            minWidth={[0, "400px"]}
+            top={[0, "32px"]}
+            zIndex={3000}
+          >
+            <DropdownMenu title="Sign in to Teleport">
+              <DropdownMenuItem
+                href="https://goteleport.com/login/"
```
Saionaro, line 63:
```suggestion
                href="/login/"
```

components/DrowpdownMenu/DropdownMenuOverlay.tsx
```diff
@@ -0,0 +1,18 @@
+import Box from "components/Box";
+
+const DropdownMenuOvrlay = () => {
```
Saionaro, line 3:
```suggestion
const DropdownMenuOverlay = () => {
```

components/DrowpdownMenu/DropdownMenuOverlay.tsx
```diff
@@ -0,0 +1,18 @@
...
+      right={0}
+      bottom={0}
+      left={0}
+      zIndex={1000}
+      background="blur(60px)"
+    />
+  );
+};
+
+export default DropdownMenuOvrlay;
```
Saionaro, line 18:
```suggestion
export default DropdownMenuOverlay;
```

components/DrowpdownMenu/DropdownMenuItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+      passthrough
+      display="block"
+      overflow="hidden"
+      px={3}
+      py={2}
+      border={["1px solid", "none"]}
+      borderColor="lightest-gray"
+      borderRadius="sm"
+      transition={transition([["background", "interaction"]])}
+      lineHeight="24px"
```
Saionaro, line 34:
```suggestion
      lineHeight="md"
```

components/DrowpdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,40 @@
+import Box from "components/Box";
+
+export interface DropdownMenuProps {
+  title: string;
+  children: React.ReactNode;
```
Saionaro, line 5: BTW, do you like these "magic" automatic react imports?

components/DrowpdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <Box
+      background="white"
+      borderRadius="default"
+      boxShadow={["none", "0 4px 40px rgba(0, 0, 0, 0.24)"]}
+      color="black"
+      overflow="hidden"
+      width={["100%", "auto"]}
+    >
+      <Box flexGrow={1}>
```
alex-kovoy, line 18: is this extra Box really required?

components/DrowpdownMenu/DropdownMenu.tsx
iAdramelk: I'm conflicted about them. On one hand for hooks I don't like them and want them to be explicit, but for types importing them explicitly also didn't feel right. Especially for events - for example `React.MouseEvent` and `MouseEvent`.

components/DrowpdownMenu/DropdownMenu.tsx
iAdramelk: Will check it, may be left from old version and is not used anymore.

iAdramelk: Checked, it was used before, because there was panel on the left in the products tab, but we removed it some time ago, it is not needed anymore.

Saionaro (approved):

PR #256, opened 2021-05-20
Fix footer links

klizhentas:


components/Footer/footer.mdx
```diff
@@ -57,7 +57,7 @@ launchpad:
       - title: (855) 818 9008
         url: tel:855-818-9008
       - title: Contact us
-        url: /contact-us/
+        url: /signup/enterprise/
```
Saionaro, line 60: Is it correct? It's is a contact us link isn't?

klizhentas: yes, as intended

PR #260, opened 2021-05-21
Add feedback component

Saionaro:
Screenshot 2021-05-21 at 17 10 23


components/Company/clients.mdx
```diff
@@ -1,259 +1,260 @@
...
-    caseStudy: /case-study/vmware-testimonial/
-  - id: volante
-    title: Volante Technologies
-  - id: wish
-    title: Wish
-  - id: xactly
-    title: Xactly Corp
-  - id: zapier
-    title: Zapier
+- id: tenxgenomics
```
iAdramelk, line 2: Why not move this list to separate json file? It will be easier to read and manage

components/Feedback/Pagination.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+export default function Pagination({
+  count,
+  onClick,
+  selected,
+  ...props
+}: Props) {
+  const dots: ReactNode[] = [];
+
+  for (let i = 0; i < count; i++) {
```
iAdramelk, line 22: Why not use map here?

iAdramelk (approved):

components/Company/clients.mdx
Saionaro: @klizhentas asked to use `md` syntax as much as possible for datasets.

components/Feedback/Pagination.tsx
Saionaro: We have only number here.
So we should use some hacky way to create a fake array and then map it.

iAdramelk: Understood, misread original code.

Saionaro: You're right btw :)
I rewrote some params to pass array.

PR #269, opened 2021-05-25
docs: Add 6.2 as latest

webvictim:
Added 6.2 to `.gitmodules` and `config.json`, then ran `yarn git-update` and did a recursive submodule checkout to update everything.

It seems to have updated a lot, not sure if this is intentional?


inertial-frame (approved):
Might be a silly question - will branch/v6 be aligned with 6.3?

webvictim:
> Might be a silly question - will branch/v6 be aligned with 6.3?

Not a silly question at all. You're absolutely right that `branch/v6` is now 6.3, but my current understanding is that we don't intend to release a 6.3, so the next release will be 7.0. @russjones and @klizhentas may know differently.

I think for now we can merge this as-is and if it becomes obvious that we're going to release 6.3, we can submit a separate PR to add it.

iAdramelk (approved):

iAdramelk:
> It seems to have updated a lot, not sure if this is intentional?

@webvictim it's half-intentional, half-not. Basically git didn't auto update submodules. We update them manually before the build on Vercel to always have latest version of docs, but we didn't commit these updates back. Because of it, it is possible that the versions of submodules in the `main` branch of this repo is quite old. So it is ok to update them together with over submodules changes.

webvictim:
Thanks @iAdramelk!

PR #273, opened 2021-05-26
BookBlock component

iAdramelk:
Desktop:

image

Mobile:

image


components/Marketo/Marketo.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { transition, StyledSystemProps } from "components/system";
+import { useMarketo } from "utils/marketo";
+
+interface MarketoProps {
+  id: string;
+}
+
+export const Marketo = ({ id }: MarketoProps) => {
+  useMarketo({ formId: id });
+  return <StyledFormFormWrapper id={`mktoForm_${id}`} />;
```
Saionaro, line 12: WDYT about hiding a wrapper while marketo loading? In that way:
```js
const loaded = useMarketo({ formId: id });
...opacity={loaded ? 1 : 0}....
```

components/BookBlock/index.ts
```diff
@@ -0,0 +1,3 @@
+export { BookBlock as default } from "./BookBlock";
+
```
Saionaro, line 2:
```suggestion
```

Saionaro (approved):

PR #279, opened 2021-05-27
Add images support in mdx and fix mdx tree bug

iAdramelk:
- Add automatic import for images in mdx props.
- Fix bug with mdx parser adding unneeded nested paragraph nodes.


utils/mdx-config.ts
```diff
@@ -18,13 +24,27 @@ export default Wrapper;
 `;
 
 interface MdxConfig {
+  rehypePlugins: PluggableList;
   remarkPlugins: PluggableList;
+
```
Saionaro, line 29:
```suggestion
```

PR #286, opened 2021-05-28
List story

Saionaro:


components/List/List.stories.tsx
```diff
@@ -0,0 +1,99 @@
+import { Story } from "@storybook/react";
+import Flex from "../Flex";
+import List, { ListItem, Props } from "./List";
```
iAdramelk, line 3: We can use `List.Item` here like in https://github.com/gravitational/next/blob/main/components/PromoPointsBlock/PromoPointsBlock.stories.tsx

iAdramelk (approved):

components/List/List.stories.tsx
Saionaro: Yeach, great idea

PR #302, opened 2021-06-09
Added pricing to nav and updated features order for enterprise card

alexwolfe:
@iAdramelk can you please review and merge these changes. Thanks!

### DIFF
--- a/components/Menu/Category.tsx
+++ b/components/Menu/Category.tsx
@@ -14,7 +14,7 @@ export interface MenuCategoryProps {
   title: string;
   description: string;
   href: string;
-  children: MenuItemProps[];
+  children?: MenuItemProps[];
 }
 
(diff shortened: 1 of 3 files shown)


components/Menu/Category.tsx
```diff
@@ -14,7 +14,7 @@ export interface MenuCategoryProps {
   title: string;
   description: string;
   href: string;
-  children: MenuItemProps[];
+  children?: MenuItemProps[];
```
iAdramelk, line 17: What is the use case for the category without items? In what situation we can have one?

alexwolfe: If you look at the branch we now have a "Pricing" button in the top nav that goes directly to the pricing page. Not all buttons in the top nav will always have dropdown menus (multiple items)

PR #306, opened 2021-06-15
Fix MDX isolation

iAdramelk:
MDX wrapper we used in `SitePage` (wrapper for all non-docs pages) was broken – it didn't supply default styles for tags like `<p>` and `<h2>`.

I fixed it's import but ts broke a bunch of other styles because of css selector complexity issue: `.wrapper h2` selector have higher complexity then `.componentHeaderClass. Because of that I reworked default element classes to only apply for tags without classes on them and fixes some problems on existing pages.


alex-kovoy (approved):
LGTM, but we need to try to simply it so overrides happen in a more clean way. Talked to Alexey and we agreed that he will create a draft PR where we can compare 2 approaches.

PR #307, opened 2021-06-15
Update Header component

iAdramelk:
New designs require that top component on the page can be placed under the header. This PR adds flag to frontmatter that allows it and corresponding logic to `<Layout>` component.


components/Logo/Logo.tsx
```diff
@@ -1,16 +1,15 @@
...
+const Logo = (props: FlexProps) => (
+  <Flex display="inline-block" lineHeight="0" {...props}>
+    <LogoSvg width="100%" height="100%" />
+  </Flex>
+);
 
-const Logo = styled(LogoSvg)<LogoProps>(space, color, position);
+Logo.defaultProps = {
+  width: "121px",
+  height: "24px",
```
alex-kovoy, line 12: you can inline these props with others, the {...props} will override them if provided. No need for extra defaultProps stuff.

components/Header/Header.tsx
```diff
@@ -25,11 +26,25 @@ const Header = ({ mode = "full" }: Props) => {
...
+      <Flex
+        as="header"
+        position="absolute"
+        top="0"
+        right="0"
+        left="0"
+        height={["48px", "80px"]}
+        alignItems="center"
+      >
+        <Centrator>
```
alex-kovoy, line 40: do we really need this extra div to center (centrator)?

alex-kovoy (approved):
with comments

components/Header/Header.tsx
iAdramelk: Its not just center things, it also maintain `max-width` and padding at different dimensions. We use this logic in most of the sections on site because we want all of them to have same max width and paddings. Ad we cant just wrap them all with it because sections has background of different colors that need to fill full page width.

E. g. see "Benefits and perks" section here https://goteleport.com/careers/ (and patterns in its background) or "Trusted by Leading Organizations" block with gray background.

components/Logo/Logo.tsx
iAdramelk: Good catch.

PR #308, opened 2021-06-15
Fix Video proportion

iAdramelk:
Video had incorrect proportions on mobile devices. This PR fixes it.


components/Video/Video.tsx
```diff
@@ -3,31 +3,45 @@ import Box, { BoxProps } from "components/Box";
...
-        title="YouTube video player"
-        frameBorder="0"
-        allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
-        allowFullScreen
-      />
+        position="relative"
+        borderRadius="md"
+        overflow="hidden"
+        pb={`${(height / width) * 100}%`}
+      >
```
alex-kovoy, line 28: I think this extra Box is unnecessary (should be possible to use 2)

alex-kovoy (approved):

components/Video/Video.tsx
iAdramelk: I tried using 2 boxes first and it works inside containers with `display: block`, but it was incorrectly calculating padding percentages inside flex containers. E. g. `padding-bottom: 52.5%` of container with `640px` width was calculated as ~`690px` inside flex container. 

My guess that this is because children of flex containers calculate their width differently from the block container ([some examples](https://stackoverflow.com/questions/23717953/padding-bottom-top-in-flexbox-layout)). With 3 containers it works fine so I decided to leave it in case we use `<Video>` inside flex container.

PR #316, opened 2021-06-17
Add SectionSteps component

iAdramelk:
Desktop:

image

Mobile:

image

Will add storybook page later.


components/SectionSteps/SectionSteps.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <Flex mt={[5, 9]}>
+          <Flex
+            width="100%"
+            justifyContent="space-between"
+            alignItems={["flex-start", "stretch"]}
+            flexDirection={["column", "row"]}
+          >
+            {children}
+          </Flex>
+        </Flex>
```
alex-kovoy, line 83: try this (seems you do not need extra div)

```tsx
<Flex
    mt={[5, 9]}>
    width="100%"
    justifyContent="space-between"
    alignItems={["flex-start", "stretch"]}
    flexDirection={["column", "row"]}
  >
    {children}
</Flex>
```

components/SectionSteps/SectionSteps.tsx
```diff
@@ -0,0 +1,89 @@
...
+    <Flex
+      flex={["0 1 auto", `0 1 ${width}px`]}
+      width="100%"
+      flexDirection="column"
+      alignItems={["center", "flex-start"]}
+      css={css({ "& + &": { ml: [0, 6], mt: [6, 0] } })}
+    >
+      <Flex alignItems="center" mb={4} maxWidth={width} flexGrow={1}>
+        <Box boxShadow="0 4px 32px rgba(0, 0, 0, 0.24)" lineHeight="0">
+          <Image src={src} width={width} height={height} layout="intrinsic" />
```
alex-kovoy, line 32: i wont be surprised if below nesting structure
```
 <Flex alignItems="center" mb={4} maxWidth={width} flexGrow={1}>
        <Box boxShadow="0 4px 32px rgba(0, 0, 0, 0.24)" lineHeight="0">
             <Image src
```             

could be replaced with 1 div
```
<Flex 
  as={Image}
  src={src} 
  width={width} 
  height={height} 
  layout="intrinsic"
  alignItems="center" 
  mb={4} 
  maxWidth={width} 
  flexGrow={1} 
  boxShadow="0 4px 32px rgba(0, 0, 0, 0.24)" 
  lineHeight="0"/>
```  

:-)

alex-kovoy (approved):
with comments

components/SectionSteps/SectionSteps.tsx
iAdramelk: Its not a real Image here. Its `next/image` component that looks something like that:

```
<div style="overflow: hidden;">
   <div ...></div> // spacer div
   <img {...props} />
</div>
```

If we try to set shadow directly on `Image` tag it will be cropped by wrapper with overflow. (And unfortunately `next/image` didn't give you api to set class to wrapper itself).

Can't combine Flex and Bot too unfortunately. Flex height can be highter that Box height. We use flex to vertically center image in the case then the image's height is smaller that height of images in other column.

components/SectionSteps/SectionSteps.tsx
iAdramelk: This one should work, yes.

PR #317, opened 2021-06-17
Add SectionTabs component

iAdramelk:
Desktop:

image

Mobile:

image

Will add storybook page later.


components/SectionTabs/SectionTabs.tsx
```diff
@@ -0,0 +1,147 @@
...
+const SectionTabsItem = ({
+  children,
+  description,
+  id,
+  selected,
+  src,
+  title,
+  onChange,
+}: SectionTabsItemProps) => {
+  const onClick = useCallback(() => onChange(id), [id]);
```
alex-kovoy, line 35: lets stop using `useCallback` as it does more harm than good when applied by default. Here is a good article on the subject 
 https://dmitripavlutin.com/dont-overuse-react-usecallback/
 
 it's OK to re-evaluate below function on this type of re-render.
 
```ts
const onClick = () => onChange(id);
```

components/SectionTabs/SectionTabs.tsx
```diff
@@ -0,0 +1,147 @@
...
+interface SectionTabsProps {
+  children: React.ReactNode;
+  subtitle: string;
+}
+
+export const SectionTabs = ({ children, subtitle }: SectionTabsProps) => {
+  const [currentTab, setCurrentTab] = useState<number>(0);
+  const onChange = useCallback((index: number) => setCurrentTab(index), []);
+
+  const childTabProps = useMemo(
```
alex-kovoy, line 99: we do not need `useMemo` here as well.

components/SectionTabs/SectionTabs.tsx
```diff
@@ -0,0 +1,147 @@
...
+};
+
+interface SectionTabsProps {
+  children: React.ReactNode;
+  subtitle: string;
+}
+
+export const SectionTabs = ({ children, subtitle }: SectionTabsProps) => {
+  const [currentTab, setCurrentTab] = useState<number>(0);
+  const onChange = useCallback((index: number) => setCurrentTab(index), []);
```
alex-kovoy, line 97: lets remove another `useCallback`

alex-kovoy (approved):
lets clean up the code by removing useMemo and useCallback and then good to merge.

components/SectionTabs/SectionTabs.tsx
iAdramelk: Fair point. Will remove them here. Price of replacing event listener in DOM here should be roughly the same.

PR #329, opened 2021-06-24
Add fallback local url to marketo

### LINKED ISSUE #327: Firefox form bug
The form does not show up on this page if Enhanced tracking is enabled on Firefox for Windows OS on this page: https://goteleport.com/enterprise-trial/

image

iAdramelk:
Fix #327

Marketo didn't work in Firefox and Safari because of tracking cookie protection. This PR adds optional proxying for the Marketo request view goteleport.com domain if the original script is blocked.

To make it work we need to add following rewrites to nginx:

- `^js/forms2/(.*)` -> `http://app-ab44.marketo.com/js/forms2/$1`
- `^form/getForm` -> `http://app-ab44.marketo.com/form/getForm`
- `^form/getKnownLead` -> `http://app-ab44.marketo.com/form/getKnownLead`
- `^index.php/leadCapture/save2` -> `http://app-ab44.marketo.com/index.php/leadCapture/save2`

CC @klizhentas ^

### DIFF
--- a/utils/marketo.ts
+++ b/utils/marketo.ts
@@ -1,22 +1,28 @@
 import { useState, useEffect } from "react";
 
+const { NEXT_PUBLIC_HOST } = process.env;
 const { NEXT_PUBLIC_MARKTO_MUNCHKIN_ID } = process.env;
 const { NEXT_PUBLIC_MARKTO_BASE_URL } = process.env;
 
-function loadScript(): Promise<boolean> {
-  return new Promise((resolve) => {
-    if (window.MktoForms2) {
-      return resolve(true);
+function loadScript(baseUrl: string): Promise<string> {
+  const url = `${baseUrl}/js/forms2/js/forms2.min.js`;
+
+  return new Promise((resolve, reject) => {
+    if (window.MktoForms2 && document.querySelector(`script[src=${url}]`)) {
+      return resolve(baseUrl);
     }
 
     const scriptElement = document.createElement("script");
 
-    scriptElement.src = `${NEXT_PUBLIC_MARKTO_BASE_URL}/js/forms2/js/forms2.min.js`;
     scriptElement.onload = () => {
       if (window.MktoForms2) {
-        return resolve(true);
+        return resolve(baseUrl);
       }
     };
+
+    scriptElement.onerror = () => reject(new Error("Can't load script"));
+
+    scriptElement.src = `${baseUrl}/js/forms2/js/forms2.min.js`;
 
     document.body.appendChild(scriptElement);
   });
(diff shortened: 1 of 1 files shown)


iAdramelk:
For details see https://blog.darwinapps.com/solution-to-embedded-marketo-forms-not-showing-up-in-firefox-c3780e6f8609

utils/marketo.ts
```diff
@@ -28,21 +34,25 @@ interface UseMarketoProps {
...
 
     window.MktoForms2.loadForm(
-      NEXT_PUBLIC_MARKTO_BASE_URL,
+      url,
       NEXT_PUBLIC_MARKTO_MUNCHKIN_ID,
       formId,
       callback
     );
-  }, [loaded, formId, callback]);
+  }, [url, formId, callback]);
```
alex-kovoy, line 55: I would remove the `callback` from the dep list as changed callback is something that regularly happens. So it should not trigger multiple `window.MktoForms2.loadForm` in this case.

If you want to guarantee that this callback is always captured by that component, you can do this:

```ts
const ref = useRef(callback);

useEffect(() => {
    if (!url) {
      loadScript(NEXT_PUBLIC_MARKTO_BASE_URL)
        .catch(() => loadScript(NEXT_PUBLIC_HOST))
        .then(setUrl)
        .catch((e) => console.error(e));

      return;
    }

    const cb = () => ref.current();

    window.MktoForms2.loadForm(
      url,
      NEXT_PUBLIC_MARKTO_MUNCHKIN_ID,
      formId,
      cb);
}, [url, formId, ref]);      
```

alex-kovoy (approved):
with comments

klizhentas (changes requested):
@iAdramelk this solution seems fragile - the moment Marketo switches the API, it would break and we would never know. What are the alternatives?

iAdramelk:
> @iAdramelk this solution seems fragile - the moment Marketo switches the API, it would break and we would never know. What are the alternatives?

@klizhentas we may create separate domain/subdomain and redirect 100% calls from it to Marketo (or use cloudfire subdomain domain for this). This way even if Marketo API changes in the future, calls will still be made.

klizhentas:
@iAdramelk I chatted to @alexwolfe and @kontsevoy. We want to switch our existing forms to Next engine and have a sync with marketo on the backend. Alex will let you know where is this on the priority list. Form requirements from me:

* Recaptcha just like on cloud form
* Marketo call on the backend with alert if it failed (and ideally retry)
* Consult with @alex-kovoy regarding security
* 
Please factor this out in a separate issue, and ask @alexwolfe about priority of this. You would need a list of forms from Alex with a spec and designs.

Issue #332, opened 2021-06-25
Using two includes in the same code block fails to render both


inertial-frame:
When I put two YAML includes into one code block only the first one renders:

```yaml
(!docs/pages/includes/app-config.yaml!)

## This section configures the 'kubernetes service'
kubernetes_service:
   ....

(!docs/pages/includes/db-config.yaml!)
```

![Screen Shot 2021-06-25 at 6 10 52 PM](https://user-images.githubusercontent.com/82530006/123493028-da7dcc00-d5e0-11eb-8322-a7ddf4457501.png)

PR #334, opened 2021-06-28
Add support for multiple includes in code block

inertial-frame:
Addresses: https://github.com/gravitational/next/issues/332 and supports: https://github.com/gravitational/teleport/pull/7421

The only place this was a problem was in `code` blocks. Using multiple code snippet includes would result in only the first getting injected and the second include path being appended as a String.

Basically, we were only checking for a single include in a block. Importantly, we update the `Node.value` every time this include check is made.

I wrapped the code block logic in a simple `match()` result array (with global flag) length-check so that the same block gets executed multiple times. Since `Node.value` gets updated every time, the contents of each include will be correctly appended. 

Doesn't impact other scenarios (using an MDX include for example).

---

Renders as desired now:

Screen Shot 2021-06-28 at 4 37 26 PM

Renders with multiple includes next to each other (via global flag):

Screen Shot 2021-06-28 at 4 37 48 PM


utils/remark-includes.ts
```diff
@@ -55,17 +59,21 @@ export default function remarkIncludes(
 
     visit<MdxastNode>(root, [hasInclude], (node, ancestors: MdxastNode[]) => {
       if (node.type === "code") {
-        const { result, error } = resolveIncludes({
-          value: node.value,
-          filePath: vfile.path,
-        });
+        const noIncludes = numIncludes(node.value);
```
iAdramelk, line 62: We only use `numIncludes` once, so we can just directly write it as `node.value.match(globalIncludeRegexp).length` here.

iAdramelk (commented):
Great work! Just one suggestion for the code.

P. S. Can you also add block with multiple includes to fixtures and run tests to make usre everything works?

utils/remark-includes.ts
inertial-frame: Yeah, I saw those tests in there - sure. I tested manually but we should probably improve the unit tests (looks like they are mostly testing path resolution right now). Happy to make the change!

Setting any length finding operation into a constant is a typical optimization technique (particularly on deeply nested objects/fields). Here, you are right to point out that it's negligible since 99% of our cases will involve single includes in a code block. Thanks!

iAdramelk: In this case you placed function creation in the constant but executing of this line `value.match(globalIncludeRegexp).length` will still happens every time the function is called with different node. So there are no difference in running this code directly VS running it inside the function. I'd even say that creating this fuctuion and executing it will take more time in this exact case. :) But the difference is probably too small to be detectable so we can disregard it.

But you are right that there are some cases when moving function creation outside has meaning. E. g. if function creation itself takes time, or if we want to preserve function instance between runs (e. g. to prevent react components rerenders) or to make code more readable.

inertial-frame: That's true. In this case, the value will decrement each time (since a `for` loop calls the second clause after each iteration (though the declared variable remains the same). As a result, moving this from the clause will actually cause a `null` value error on the first file that's checked!

Sorry, forgot the other reason I left this like that!

inertial-frame:
@iAdramelk - please see my comment about the use of `const`. I also updated the tests:

1. Renamed one of the includes to `includes-table.mdx`
2. Added a test for multiple includes within a code block (changed by this PR) and multiple includes in a text block (unchanged by this PR).

utils/fixtures/includes-multiple-source.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+Code block with multiple includes.
+
+```css
+.example1 {
+  width: 50%;
+  height: 50%;
+}
+
+(!include.css!)
```
inertial-frame, line 11: Note the whitespaces EOF.

utils/remark-includes.test.ts
```diff
@@ -23,7 +23,7 @@ const transformer = (
     .use(remarkIncludes, pluginOptions)
     .processSync(vfile(vfileOptions));
 
-describe("utils/remark-variables", () => {
+describe("utils/remark-includes", () => {
```
inertial-frame, line 26: Assuming this was copied from the other tests?

iAdramelk: Thanks

iAdramelk (approved):
LGTM

inertial-frame:
Merging with approval from @iAdramelk

Issue #251, opened 2021-05-20
Review use of the "Edit" button on the website for user edits


pschisa:
### Problem Statement
Currently on the website, a user can select the "edit" button to start an automatic workflow to open a Github PR for documentation edits.  I have tried a few various docs edits through the current workflow and noticed several problems that would prevent a user change/PR from being useful:

1) The edit button by default will open the PR against the current branch version of the docs (example: v6.1), which will not allow for the changes to persist after the docs version is updated outside of that branch.

2) Back/forward ports that are generally required to get docs changes in all the right spots are not exposed to the end user.  There is currently no automation system to make the required ports off a single docs edit making this a manual process, adding a further burden to the end user (which may not be familiar with Github if using the docs edit link).

3) Documentation changes require multiple code owner approvals and getting these approvals generally requires advocating for the PR to be reviewed internally.  End users do not have the same access to advocate for their PRs to be reviewed in a timely manner

4) The edit button does not walk users through any sort of testing/validation of the documentation changes as they will appear on the Website.  This makes it difficult for end users to be sure their changes are going look the way they intended.

All these issues combine to make the edit button essentially non-functional for affecting real Teleport documentation changes from an end user perspective via the edit button.

### Suggestions
I think the easiest solution is to change the edit button from opening a PR to opening a documentation issue (possibly with some auto-generated format based on the docs page selected).  This will allow users to make the Teleport team aware of documentation issues found without the end user having to worry about how the change needs to be implemented in our docs system.  Then, if the user is comfortable they could attempt to make a PR and associate it with the issue.  The issue could then be used to track required back/forward ports as well, which would likely be performed by a Teleport engineer anyway.

Otherwise, if we continue using the edit button to open PRs, I would recommend the following steps need to be taken to make the edit button more viable:
1) Open against master branch always so that the changes are at least included in all future major releases without requiring porting
2) Implement automation that will make the required forward/back ports off the single change to master made by the end user
3) Find a method to ensure documentation PRs are reviewed regularly without requiring an internal advocate for review
4) Find a method/process so that end users can easily test their docs changes locally before committing their PR

Issue #292, opened 2021-06-01
Ratings or ranking system


inertial-frame:
I've had a few conversations with people that have alluded to this kind of idea.

Basically, it'd be great to have some kind of direct feedback input:

1. I don't think readers should have to log in. The feedback should not be tethered to a user account. This vastly simplifies the engineering requirements and also improves the reader experience.

2. We need to think carefully about what the feedback metrics should be. Most companies use something like Customer Satisfaction, Like / Dislike the documentation, along with a suggestion feature. 

We currently direct people to create an issue and/or to our community Slack channel. One thing I like about our current system is that we have a unified suggestion pipeline. We should route people to our Issues Page AND note the presence of our other support channels. 

We could go also further and unite those two into a Slack webhook that links to the Issue in GitHub.

3. So, text feedback inputs could be a field that results in the creation of a ticket and Slack webhook or just what we have today (I think both are fine). The main focus here would be on keeping and tracking whatever rating system we'd implement.

Relevant metrics:

I. Page
II. Time/Date
III. Rating

inertial-frame:
Thinking a bit more about this:

1. Probably could be implemented as a simple MDX component that would be added to the bottom of every page (manually). The advantage of this approach is that we can be selective about which pages we want this on. (Maybe not landing pages.)
2. Pages would be identified with a UUID - we can update this or change this as needed (say after a major page update). This helps to track a page when it's renamed, refactored, moved, etc.
3. Probably a simple table:

| UUID | Int | Int |
| --- | --- | --- |
| afefwe | 100 | 50 |

4. Maybe use AWS Lambda (or some equivalent like Azure Functions) for a simple RESTful API.
5. The MDX component would have an Up and Down button that would make a POST. 
6. The POST operation would increment the relevant column and would return that updated value which would be used to populate the field. (Rerender)
7. ComponentDidMount executes a GET to populate the values.

inertial-frame:
We could get a little more precise by adding:

1. In addition to the UUID, a Version prop to the MDX component.
2. Adding a second Table that would track time, UUID, vote, and version to give a better time-based metric.

benarent:
Plan is create a small Lambda service that'll receive. Send data to Postgres and display report in https://metabase.asteroid-moon.teleport.sh/ 

- Page Name
- Version
- rating ( 1 / 0 ) 
- timestamp
- browser, OS

PR #341, opened 2021-06-29
Update edit button link

inertial-frame:
URL should create a documentation issue rather than a code edit/branch/PR.

- It will generate a URL like the following: https://github.com/gravitational/teleport/issues/new?assignees=&labels=documentation&template=documentation.md&title=[v.6.0]%20/docs/pages/cloud.mdx
- It will populate the Issue title field with the name of the file and version number.
- Addresses: https://github.com/gravitational/next/issues/251

### DIFF
--- a/utils/docs-helpers.ts
+++ b/utils/docs-helpers.ts
@@ -35,7 +35,7 @@ export const getVersionConfigPath = (filepath: string) => {
 export const getGithubURL = (filepath: string) => {
   const current = getVersion(filepath);
   const root = getVersionRootPath(filepath);
-  const ghIssuePath = `${NEXT_PUBLIC_GITHUB_DOCS}/issues/new?assignees=&labels=documentation&template=documentation.md`;
+  const ghIssuePath = `${NEXT_PUBLIC_GITHUB_DOCS}/issues/new?labels=documentation&template=documentation.md`;
 
   return branches[current]
     ? `${ghIssuePath}&title=[v.${current}]%20${filepath.replace(root, "")}`
--- a/utils/docs-helpers.ts
+++ b/utils/docs-helpers.ts
@@ -33,7 +33,7 @@ export const getVersionConfigPath = (filepath: string) => {
 };
 
 export const getGithubURL = (filepath: string) => {
-  const current = getVersion(filepath); 
+  const current = getVersion(filepath);
   const root = getVersionRootPath(filepath);
   const ghIssuePath = `${NEXT_PUBLIC_GITHUB_DOCS}/issues/new?assignees=&labels=documentation&template=documentation.md`;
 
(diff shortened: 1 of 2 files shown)


pschisa:
I like it.  A couple of questions I was thinking about:

1) Should the auto-generated title field also contain something unique to the request so that multiple docs issues against the same page don't end up with the same name?

2) Should we update the website's `Edit` button as well?  It seems like something more like `Report Error` or `Report Issue` would be more accurate now.

utils/docs-helpers.ts
```diff
@@ -35,11 +35,9 @@ export const getVersionConfigPath = (filepath: string) => {
 export const getGithubURL = (filepath: string) => {
   const current = getVersion(filepath);
   const root = getVersionRootPath(filepath);
+  const ghIssuePath = `${NEXT_PUBLIC_GITHUB_DOCS}/issues/new?assignees=&labels=documentation&template=documentation.md`;
```
iAdramelk, line 38: Nitpick. 

```suggestion
  const ghIssuePath = `${NEXT_PUBLIC_GITHUB_DOCS}/issues/new?labels=documentation&template=documentation.md`;
```

I think we can omit assignees part here.

iAdramelk (approved):

utils/docs-helpers.ts
inertial-frame: Done. Thanks!

inertial-frame:
@pschisa - thanks for the great suggestions (and creating the well-thought-out issue):

1. I've tweaked the button to `Improve` since there's a fixed-width (anything larger than that will be cropped). I think this is an improvement that meets the spirit of your suggestion. What do you think:

Screen Shot 2021-07-01 at 11 05 11 AM

2. I think having the file name and the version will create a fairly unique namespace. Most pages don't see more than 1 or 2 PR's at a time. Also, not sure how we could make that more specific at this point (although I agree, that'd be awesome.) We basically generate the URL, title, etc. from the file path and predefined GitHub repo URL.

I think we should also include a Slack Org/Channel link in the Documentation Template on GitHub so customers can be directed into that support channel directly who need it.

That being said, there's a more comprehensive Rating System that's planned to also help us track improvements: https://github.com/gravitational/next/issues/292 - I think this change and a dedicated documentation rating system will help us out a lot.

iAdramelk:
@inertial-frame you can change the button width on this line https://github.com/gravitational/next/blob/690aadd149a473a6c828440be5ff51460ff4115c/components/DocsPage/Header.tsx#L80 , but please check that it have enough space on mobile to show them on one line.

inertial-frame:
> @inertial-frame you can change the button width on this line
> 
> https://github.com/gravitational/next/blob/690aadd149a473a6c828440be5ff51460ff4115c/components/DocsPage/Header.tsx#L80
> 
> , but please check that it have enough space on mobile to show them on one line.

Saw this but didn't want to modify the default settings. Thanks!

inertial-frame:
Merging with approval and sign-off from @iAdramelk

Issue #305, opened 2021-06-14
Some tables used in includes render incorrectly


inertial-frame:
If I put the following in a MDX include:

```
Teleport services listen on several ports. This table shows the default port numbers.

| Port | Service | Description | Ingress | Egress |
| --- | --- | --- | --- | --- |
| `443` | Proxy | Default proxy port when using ACME for TLS/SSL. | {/* TODO */} | {/* TODO */} |
| `3022` | Node | SSH port to the Node Service. This is Teleport's equivalent of port `22` for SSH. | Allow inbound traffic from the proxy host. | Allow outbound traffic to the proxy host. |
| `3023` | Proxy | SSH port clients connect to after authentication. A proxy will forward this connection to port `3022` on the destination node. | Allow inbound traffic from SSH clients. | Allow outbound traffic to SSH clients. |
| `3024` | Proxy | SSH port used to create "reverse SSH tunnels" from behind-firewall environments into a trusted proxy server. | {/* TODO */} | {/* TODO */} |
| `3025` | Auth | SSH port used by the Auth Service to serve its Auth API to other nodes in a cluster. | Allow inbound connections from all cluster nodes. | Allow outbound traffic to cluster nodes. |
| `3026` | Kubernetes | HTTPS Kubernetes proxy `proxy_service.kube_listen_addr`. Port used for `kubectl` to access Teleport. | {/* TODO */} | {/* TODO */} |
| `3027` | Kubernetes | Kubernetes Service `kubernetes_service.listen_addr` | {/* TODO */} | {/* TODO */} |
| `3036` | DB | Teleport MySQL proxy listen address. | {/* TODO */} | {/* TODO */} |
| `3080` | Proxy | HTTPS port clients connect to. Used to authenticate `tsh` users and web users into the cluster. | Allow inbound connections from HTTP and SSH clients. | Allow outbound connections to HTTP and SSH clients. |

```

And include it like so:

```
(!docs/pages/includes/default-ports.mdx!)
```

It will render each line of the include as a String (`P` HTML element to be more precise) rather than as an HTML Table element.

![Screen Shot 2021-06-14 at 3 57 21 PM](https://user-images.githubusercontent.com/82530006/121958574-711cd400-cd29-11eb-8e22-8b44e269e77d.png)

(The first table in the image above was added just to test that it wasn't due to something in the second table)

inertial-frame:
Follow-up:

[...]

PR #351, opened 2021-07-15
Fix/included table render

Alqanar:
Fix https://github.com/gravitational/next/issues/305

Corrected the use of including tables.
At the moment markdown tables work fine in any form with any contents, html tables with empty cells do not work correctly yet because of error in mdx parsing


iAdramelk (approved):
LGTM

PR #352, opened 2021-07-15
Cole/features

### LINKED ISSUE #348: New Features Page
We are redesigning the features page (see image below). The new update will include the following: 
- Complete redesign of the current features page: https://goteleport.com/features/
- Update to the top "Products" nav, we'll be adding a "Teleport features" as one of the options. I'll provide an update design soon. 

**Current Page:** https://goteleport.com/features/ (useful for links) 
Avocode Demo: https://app.avocode.com/view/dec5ab183ae94c649391a26d5fb56181/_/inspect

image

C-STYR:
fixes #348


pages/features/index.mdx
```diff
@@ -0,0 +1,198 @@
...
+  mode={meta.layout}
+  subtitle="Access Plane"
+  title="Teleport Features"
+  bg="wave"
+  description={
+    <>Nearly all Teleport features are available in the open source package.</>
+  }
+></SectionHeader>
+<Section bg="flatGray">
+  <Flex flexDirection="row" pl={[3, 4, 11]} pt={[4, 7, 9]} pb={[4, 6]}>
```
iAdramelk, line 36: We use reusable components as much as we want. So to make tiles block reusable we should probably create separate component for tiles' group with `icon`, `title` and `children` fields. It will be also easier to use on this page, because it is already repeated twice.

pages/features/index.mdx
```diff
@@ -0,0 +1,198 @@
...
+    <Icon name="apartment" size="lg" fill="#512FC9" />
+    <Box fontSize="header-3" color="darkest" lineHeight="lg" pl={3}>
+      Enterprise Features
+    </Box>
+  </Flex>
+  <Flex px={[3, 3, 10]} pb={6}>
+    <FeatureTileDisplay>
+      <FeatureTile
```
iAdramelk, line 44: This component is used for features on this page, but it can be used for something also on the other page, so I'd rename to something more neutral.

 Also instead of exporting it separately, we better export is as a part of the parent component and in case of `SectionTabs`.

pages/features/index.mdx
```diff
@@ -0,0 +1,198 @@
...
+  }
+></SectionHeader>
+<Section bg="flatGray">
+  <Flex flexDirection="row" pl={[3, 4, 11]} pt={[4, 7, 9]} pb={[4, 6]}>
+    <Icon name="apartment" size="lg" fill="#512FC9" />
+    <Box fontSize="header-3" color="darkest" lineHeight="lg" pl={3}>
+      Enterprise Features
+    </Box>
+  </Flex>
+  <Flex px={[3, 3, 10]} pb={6}>
```
iAdramelk, line 42: Right now features take 100% of the screen, but we probably want to limit maximum with of the contetn, or it will become unreadable on the larger screens. You can use `layout/Cetrator` component for this it will limit max width and center content for sections. You can check examples of its usage in `Section*` component.

pages/features/index.mdx
```diff
@@ -0,0 +1,198 @@
...
+  subtitle="Access Plane"
+  title="Teleport Features"
+  bg="wave"
+  description={
+    <>Nearly all Teleport features are available in the open source package.</>
+  }
+></SectionHeader>
+<Section bg="flatGray">
+  <Flex flexDirection="row" pl={[3, 4, 11]} pt={[4, 7, 9]} pb={[4, 6]}>
+    <Icon name="apartment" size="lg" fill="#512FC9" />
```
iAdramelk, line 37: Let's try to use `color` here. If it still doesn't work after moving to `tsx` component, please let me know and we will debug it together.

pages/features/index.mdx
```diff
@@ -0,0 +1,198 @@
...
+        title="IoT and Edge Connectivity"
+        href="https://goteleport.com/teleport/how-it-works/ssh-kubernetes-access-on-the-edge/"
+      >
+        Teleport provides access to SSH servers and Kubernetes clusters anywhere
+        in the world, behind NAT. From restaurants to autonomous vehicles, no
+        VPN required.
+      </FeatureTile>
+    </FeatureTileDisplay>
+  </Flex>
+  <Flex flexDirection="column" alignItems="flex-end" px={[4, 5, 11]} pt={6}>
```
iAdramelk, line 180: We better move it to separate component too. Also we probably should discuss with Alex Wolfe hot to make it more universal so we can use it not only on this page, but also on the others.

pages/features/index.mdx
```diff
@@ -0,0 +1,198 @@
...
+    <Box fontSize="text-xl">
+      <h1 color="darkest" fontSize="header-1" >How Teleport Works</h1>
```
iAdramelk, line 182: Page can have only 1 `h1` header. So let's switch it to `h2`. Also `styled-system` will not work with `<h1>` directly, to make it works we need to use `<Box as="h2"></Box>`.

pages/features/index.mdx
```diff
@@ -0,0 +1,198 @@
...
+      Teleport is a single binary which enables secure access to SSH nodes, Kubernetes
+      clusters, web apps, PostgreSQL and MySQL databases behind NAT. Teleport is trivial
+      to setup as a Linux daemon or in a Kubernetes pod. <Link color="light-blue" href="/">Learn more about how Teleport
+      works.</Link>
+    </Box>
+    <Box width="103%" my={[1, 4]} >
+      <NextImage
+        src="./assets/How-it-works.png"
```
iAdramelk, line 190: Image is blurry on the actual site now.  As for the section itself, lets try to use `<Centrator>` too. We probably want to limit it's max width with absolute value that is equal to the real image dimensions and center it.

pages/features/index.mdx
```diff
@@ -0,0 +1,198 @@
...
+  Section,
+  Flex,
+  Box,
+  Icon,
+  TryTeleport,
+  Link
+} from "components";
+
+import FeatureTile from "components/Tile/FeatureTile";
+import FeatureTileDisplay from "components/Tile/FeatureTileDisplay";
```
iAdramelk, line 22: to make it easier to use we want to reduce number of import sources. In the perfect world we want to import all components from `components`.

components/Tile/FeatureTileDisplay.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+type FeatureTileType = React.ReactElement<typeof FeatureTile>;
+
+export interface FeatureTileProps {
+  children: FeatureTileType;
+}
+
+const FeatureTileDisplay = ({ children }: FeatureTileProps) => {
+  return (
+    <Flex
```
iAdramelk, line 12: We probably want to use `import { Centrator } from "components/layout"` here to limit max width of the component. And also move section header here too (see comment in the `page/features` for details).

Also we probably want to move wrapper itself to it's own folder instead of `Tile`.

components/Tile/FeatureTile.tsx
```diff
@@ -0,0 +1,81 @@
...
+    borderRadius: "md",
+    textDecoration: "none",
+    boxShadow: "0 1px 4px rgba(0, 0, 0, 0.24)",
+    mb: 4,
+    mx: 2,
+    bg: "white",
+  })
+);
+
+const StyledButton = styled(Button)({
```
iAdramelk, line 75: Lets discuss with Alex how often we want to use this styles. We want consistent button styles across the site. So should either leave them as they are in `Button` already, or add new variation to the `Button`component itself.

components/Tile/FeatureTile.tsx
```diff
@@ -0,0 +1,81 @@
...
+    </StyledWrapper>
+  );
+};
+
+const StyledWrapper = styled(Flex)(
+  css({
+    flexDirection: "column",
+    maxWidth: "calc((100% - 80px) / 4)",
+    minWidth: ["340px", "280px", "180px"],
+    maxHeight: "500px",
```
iAdramelk, line 65: Why do we need fixed `minWidth` and fixed `minHeight` here? Technically we only want to have some fixed number of elements in the line and for them to have the same height.

We can do first by changing  `calc((100% - 80px) / 4)` this line to something like `calc((100% - 60px) / 3)` (60px should be replaced width the sum of gsaps between elements).

Second should automatically work with `alignItems: stretch`.

components/Tile/FeatureTile.tsx
```diff
@@ -0,0 +1,81 @@
...
+      >
+        <NextImage src={src} alt={`${title}`} height={193} width={282} />
+      </Flex>
+
+      {/* bottom half */}
+      <Flex flexDirection="column" pt={3} px={4} borderRadius="0 0 8px 8px">
+        <Box fontSize="text-md" lineHeight="md" color="black" fontWeight="bold">
+          {title}
+        </Box>
+        <Box fontSize="13px" lineHeight="md" color="gray" pb={3}>
```
iAdramelk, line 47: Let's replace it with the value form the `theme.ts`, we don't want to add rendom sizes here. And let's discuss it with Alex if needed.

components/Tile/FeatureTile.tsx
```diff
@@ -0,0 +1,81 @@
...
+          {children}
+        </Box>
+      </Flex>
+      <Flex flexDirection="column-reverse" px={4} pb={4} height="100%">
```
iAdramelk, line 51:
```suggestion
      <Flex flexDirection="column-reverse" px={4} pb={4} flexGrow={1}>
```

components/Tile/FeatureTile.tsx
```diff
@@ -0,0 +1,81 @@
...
+        flexDirection="column"
+        alignItems="center"
+        py={3}
+        px={4}
+        backgroundImage={`url("${backgroundImage}")`}
+        backgroundPosition="center"
+        backgroundSize="cover"
+        borderRadius="8px 8px 0 0"
+      >
+        <NextImage src={src} alt={`${title}`} height={193} width={282} />
```
iAdramelk, line 39: Are we sure that all images we use in this component will have these sizes?

components/Tile/FeatureTile.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+const FeatureTile = ({
+  children,
+  src,
+  title,
+  feature,
+  href,
+}: FeatureTileProps) => {
+  const backgroundImage = feature === "enterprise" ? enterpriseBG : coreBG;
```
iAdramelk, line 25: Let's make this prop name more neutral. We may want to use this component not only for the features.

components/Section/Section.tsx
```diff
@@ -28,6 +34,14 @@ const getBG = (color: BGColor) => {
         backgroundColor: "white",
         backgroundImage: `url(${squaresBG})`,
       };
+    case "flatGray":
+      return {
+        backgroundColor: "#f6f8f9",
```
iAdramelk, line 39:
```suggestion
        backgroundColor: "page-bg",
```

components/Section/Section.tsx
```diff
@@ -28,6 +34,14 @@ const getBG = (color: BGColor) => {
...
+      };
+    case "flatWhite":
+      return {
+        backgroundColor: "#ffffff",
```
iAdramelk, line 43: We don't need this one. It's a default value (see below).

components/Icon/svg/circleCheck.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generated by IcoMoon.io -->
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="20" height="20" viewBox="0 0 20 20">
+<path d="M9.5 20c-2.538 0-4.923-0.988-6.718-2.782s-2.782-4.18-2.782-6.717c0-2.538 0.988-4.923 2.782-6.718s4.18-2.783 6.718-2.783c2.538 0 4.923 0.988 6.718 2.783s2.782 4.18 2.782 6.718-0.988 4.923-2.782 6.717c-1.794 1.794-4.18 2.782-6.718 2.782zM9.5 2c-4.687 0-8.5 3.813-8.5 8.5s3.813 8.5 8.5 8.5 8.5-3.813 8.5-8.5-3.813-8.5-8.5-8.5z" fill="#000000"></path>
```
iAdramelk, line 5: Let's change both `fill` values to `currentColor` from `#000000`.

components/Icon/Icon.tsx
```diff
@@ -27,7 +28,7 @@ const Icon = ({ name, size = "md", ...props }: IconProps) => {
 
   return (
     <Box size={sizes[size]} {...props}>
-      <IconSVG width="100%" height="100%" display="block" />
+      <IconSVG width="100%" height="100%" display="block" fill={fill} />
```
iAdramelk, line 31: We don't need `fill` here if we use `currentColor` in the icons.

components/Header/Header.tsx
```diff
@@ -11,58 +11,71 @@ import { Centrator } from "components/Layout";
...
-            height={["24px", "30px"]}
-            color="dark-purple"
-          />
-        </Centrator>
-      </Flex>
-    );
-  }
+  switch (mode) {
+    case "none":
+      return <></>;
```
iAdramelk, line 33: You can return `''` or `false` or `undefuned` here. it will render correctly.

iAdramelk (changes requested):
Let's rename images to follow the same naming convention. Right now we have: `How-it-works.png`, `audit-log.png ` and `enterpriseBG.png` at the same time.

I use `lower-case-kebab-case` then possible.

components/Header/Header.tsx
```diff
@@ -21,13 +21,13 @@ interface HeaderProps {
...
-  const toggleNavigaton = useCallback(
-    () => setIsNavigationVisible((value) => !value),
-    []
-  );
+  const [isNavigationVisible, setIsNavigationVisible] =
+    useState<boolean>(false);
+  const toggleNavigaton = useCallback(() => {
+    setIsNavigationVisible((value) => !value);
+    blockBodyScroll(isNavigationVisible);
+  }, [isNavigationVisible]);
```
alex-kovoy, line 29: This is not a good case for using `useCallback`, a simple function would do.

```js
 const toggleNavigaton = () => {
    setIsNavigationVisible(!isNavigationVisible);
    blockBodyScroll(isNavigationVisible);
};
```  

Here is a nice article on the subject.
https://dmitripavlutin.com/dont-overuse-react-usecallback/

C-STYR: @alex-kovoy  thanks for the review--just to let you know, the code you have commented on exists within this PR only as an artefact of merge conflict resolution with PR #350; I am not in fact the author of the useCallback code block.

Please see here: [PR #350 ](https://github.com/gravitational/next/pull/350/files)

I will however enjoy the article you posted, thanks.

pages/features/index.mdx
C-STYR: Alex wanted the image aligned left, not centered, but I'll discuss with him.

components/Section/Section.tsx
C-STYR: The default value also has a wave image overlay, which we would then need to be removed on a case-by-case basis, unless I'm misunderstanding?

iAdramelk: You are right, I missed them, because they were not shown in PR>

components/Header/Header.tsx
C-STYR: This seems stylistic...is there a performance or other benefit, or just conventional?

iAdramelk: It's more conventional, `<></>` will add additional import and function call, but its performance impact will probably be negligible. You can disregard it if you want.

components/Tile/FeatureTile.tsx
C-STYR: Per Alex, it's going to stay at 13px and we're going to add it to theme.ts

components/Tile/FeatureTile.tsx
C-STYR: Interesting, you'll have to explain your preference to me.  I would have thought the two were functionally equivalent.

iAdramelk: `height: 100%` is calculated form something. Usually form the height of the parent container. It will work in a lot of cases, but if we have other children with some height or if their height is calculated in % too it may not. `flexGrow: 1` in this case is sayng that container should take all "free" space in the container (its acutually more comlpex because multiple elements can have `flexGrow` and `flegGrow` value may be more that `1`, but in this case it is like this).

PR #353, opened 2021-07-20
Add scope support

iAdramelk:
Add 3 possible scopes for the docs page: `oss`, `enterprise`, `cloud`.

Switcher in the head of docs pages changes scopes, scope is persistent between page navigation. Selecting the `cloud` scope will automatically redirect to the root of the current version of the docs if user is not on it right now. 

Scope can also be changed buy adding `?scope=cloud` to the page url if we want to target exact scope, but adding `cloud` to the non-current version will cause redirect to `/docs/`.

There are three components that are affected by scope:

1. Tabs.

   TabItem now have an optional `scope` field (either string or array of strings). On scope change the tab with associated scope will be selected if possible. Tabs still can be changed manually, it will not change global scope.

   If the version is not current tabs associated with the `cloud` scope will show warning and link to the current version instead of actual content.

2. Details.

   It's a new component to show/hide some additional content. It can be used without scope too.

   If it have associated scope (or array of scopes), it will be opened on scope change. If scope is not current, or if there are no scope, it will be collapsed.

   If `scopeOnly` field is set to `true`, component will be rendred only if the scope is selected and fully hidden otherwise.

   If the collapsible scope is `cloud` it will not be rendered at all in the non-current versions.

   `min` is an optional field to add info about the minimum version of the `teleport`. It will be displayed in the head of the component.

3. Notice

   Notice now have an optional `scope` field (either string or array of strings). If the filed is set, component will only be rendered if the associated scope is selected, otherwise it will be hidden.

P. S. Wasn't able to add examples to the docs pages, because they were in different repo. So to test it it is required to checkout the code locally and change docs pages manually with something like that:

```
<Notice title="Notice title" type="danger" scope={["oss", "enterprise"]}>
  Notice text
</Notice>

<Details
  title="Some very very very very very very long title that we need to crop"
  opened={true}
  scope={["cloud", "enterprise"]}
  scopeOnly={true}
  min="7.0"
>
  Some content
</Details>

<Tabs>
  <TabItem label="One">Content 1</TabItem>
  <TabItem label="Two">Content 2</TabItem>
  <TabItem label="Three" scope={["oss", "enterprise"]}>Content 3</TabItem>
</Tabs>
```


Alqanar (approved):
LGTM

klizhentas (approved):

Issue #166, opened 2021-04-02
Code snippet


klizhentas:
## Description

Code snippet should add `$` to all code automatically. Remove all `$` from all code snippets.

inertial-frame:
As a follow-up to this - is it possible to both add the `$`  symbol before code snippets but maybe wrap it in a separate element from the block that gets copied on selection?

Personally, I prefer having `$` depicted/rendered (stresses non-root access, helps clarify lines/commands) but not having one copied (since it's not always appropriate for the terminal used, user type, etc.).

I think @benarent requested this feature too: https://github.com/gravitational/next/issues/96

sshahcodes:
Adding `$` automatically will not be an easy task since it's very hard to predict when and where to add it. For example in a multi-line code block like below, how can we decide that `$` should be placed before specific line? 
```
# Create local config and data directories for teleport, which will be mounted into the container
mkdir -p ~/teleport/config ~/teleport/data

# Generate a sample teleport config and write it to the local config directory.
# This container will write the config and immediately exit - this is expected.
docker run --hostname localhost --rm \
  --entrypoint=/bin/sh \
  -v ~/teleport/config:/etc/teleport \
  quay.io/gravitational/teleport:6 -c "teleport configure > /etc/teleport/teleport.yaml"

```

As mentioned in #96 which references Heroku sample, I looked into Docker docs website and it does the same. i.e. makes `$` unselectable using `user-select: none`. If you look at the [docker docs source file](https://raw.githubusercontent.com/docker/docker.github.io/master/engine/install/ubuntu.md) , they do have `$` in `.md` files. So they must be using some transformer to make it unselectable while showing in the page. They do not offer copy to clipboard feature though!. Neither does Heroku. 


Currently, in our case, code selection can be done in two ways:
1. Directly selecting text using keyboard/mouse text select: `user-select: none` works for this case but we will need a transformer to inject CSS for `$` character.  
2. Using copy to clipboard button (rendered inside code block): making `$` character unselectable using `user-select: none` will not work here because it is copied from `codeRef.current.innerText` of HTMLDivElement (css feature will not have effect on HTML inner dom)

So we have two possible solutions:
1. [ ] Build a `.md` transformer plugin that will render `$` as `user-select: none` element. This will solve the first case described above.
2. [ ] Manually filter out `$` character from copied content when using the Copy button. This will solve the second case described above.

[...]

Issue #289, opened 2021-05-31
Placeholder text for user values


inertial-frame:
Google now uses really cool placeholder styling in their documentation: https://cloud.google.com/anthos/clusters/docs/on-prem/1.5/how-to/ssh-cluster-node

This helps create a unified experience and prevents copy-pasting with values that should be customized or specific for the reader.

Is there a way to add a subcomponent that can be added into a code snippet that accomplishes that?

Issue #275, opened 2021-05-26
Code snippet component


klizhentas:
Consider code snippet component:

```
<Code syntax="go">
   <Command> tctl ls </Command>
   Any command output
</Code>
```

This code snippet should render as:

```
$ tctl ls 
 Any command output
```

The copy selection behavior should only copy `tctl ls` and add `$` to the command. So copying this will copy only the command. Combine it with #166 and #84

inertial-frame:
Question: is it better to render the code block like we currently do (`bash`, white text on black background) but then render the console output differently?

1. Docker docs don't seem to support the handy copy-paste functionality ours do but I really like how they use color to contrast the console outputs from the command text:

<img width="856" alt="Screen Shot 2021-05-27 at 6 52 17 PM" src="https://user-images.githubusercontent.com/82530006/119910512-e7919780-bf1c-11eb-8fcc-6d8ddfad891a.png">

2. Kubernetes, on the other hand, puts the output in a follow-up block:

<img width="754" alt="Screen Shot 2021-05-27 at 6 51 44 PM" src="https://user-images.githubusercontent.com/82530006/119910497-dfd1f300-bf1c-11eb-9eb0-b342066386e5.png">

inertial-frame:
3. One last related question: https://github.com/gravitational/next/issues/289 - can we also add support for a child element or prop that would allow a highlighted or bolded placeholder value to be emphasized?

Issue #84, opened 2021-02-01
In page edits


klizhentas:
## In page edits

When following a code guide that spans several blocks, it's easy to loose context of what the parameters are.

Google solves this using editable code widgets that distribute to other parts of the snippet. 

https://cloud.google.com/dns/docs/zones

![widgets](https://user-images.githubusercontent.com/13717/102041789-31c6f800-3d85-11eb-9ce4-c38f8213fa86.png)

inertial-frame:
Related: https://github.com/gravitational/next/issues/289 <- styling difference.

PR #358, opened 2021-07-26
filter '$' character from copied code text from docs pages

sshahcodes:
A character filter function is implemented which removes a given character from a text line if it is the first character in that line. This should solve removing `$` character from code texts when copied from the copy button (2nd problem case as mentioned in https://github.com/gravitational/next/issues/166#issuecomment-886538864).

### DIFF
--- a/components/MDX/Pre.tsx
+++ b/components/MDX/Pre.tsx
@@ -7,6 +7,8 @@ import Code from "components/Code";
 import Icon from "components/Icon";
 import HeadlessButton from "components/HeadlessButton";
 
+import DeselectChar from "../../utils/deselect-character";
+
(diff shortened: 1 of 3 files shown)


iAdramelk:
@sshahcodes thank you very much for you hard work! Unfortunately our issues are not liked together as they should. We planned to solve this together with https://github.com/gravitational/next/issues/275 which add new syntax to code blocks, so I can't merge this one as is, because it will break this new widget then we add it. But I will try to port this pr to this new widget and use it with it.

sshahcodes:
@iAdramelk I agree. There is actually #275, In-page edits #84, and #289 which also should be solved within a single code rendering update.

Issue #238, opened 2021-05-12
Placeholder or blurred image for images on page load


inertial-frame:
Noticed that some of the images in our docs take a bit to render on the page - there's maybe a .5 second delay between images rendering and the rest of the React component mounting.

I feel like this breaks my flow as a reader and wonder about the impact on bounce rates.

Can we add something like https://novelist.xyz/tech/progressive-background-image-loading/ - I have a very rudimentary example of this here: https://codepen.io/thoughtscript/pen/wvaROar

inertial-frame:
Follow-up: wondering if some of this could be improved by tweaking our CDN settings. Verifying if we're self-serving.

PR #361, opened 2021-07-27
Switched to next/image component

Alqanar:
https://github.com/gravitational/next/issues/238


components/Card/Card.tsx
```diff
@@ -4,13 +4,14 @@ import NextImage from "next/image";
 import Box from "components/Box";
 import Flex from "components/Flex";
 import Button from "components/Button";
+import { NextImageType } from "common-types/next-image";
```
iAdramelk, line 7: `next-env.d.ts`

components/Card/Card.tsx
```diff
@@ -20,11 +21,11 @@ const cardBackgrounds = {
   core,
 };
 
-const Card = ({ children, src, title, cardBG, href }: CardProps) => {
+const Card = ({ children, img, title, cardBG, href }: CardProps) => {
```
iAdramelk, line 24: I'd change it back for now and then fix it with the remark-plugin instead.

components/Card/Card.tsx
```diff
@@ -20,11 +21,11 @@ const cardBackgrounds = {
...
   //add additional backgrounds to cardBackgrounds
   const backgroundImage = cardBackgrounds[cardBG];
   return (
-    <StyledWrapper>
+    <StyledWrapper as="li">
```
iAdramelk, line 28: It can be used without wrapper, so it will not always be inside `ul`.

components/Card/CardDisplay.tsx
```diff
@@ -40,13 +40,15 @@ export function CardDisplay({
           </Flex>
           {/** Card display logic */}
           <Flex
+            as="ul"
```
iAdramelk, line 43: Cards are not always used inside this wrapper, so I'd live it as `div`.

iAdramelk (changes requested):

PR #363, opened 2021-07-28
Cole/product banner

C-STYR:
Created ProductBanner component for use in Access family of pages;
Screen Shot 2021-07-28 at 1 12 01 PM


components/ProductBanner/ProductBanner.tsx
```diff
@@ -0,0 +1,150 @@
...
+        mb={2}
+      >
+        {title}
+      </Box>
+      <Box
+        alignItems="flex-start"
+        fontSize="text-md"
+        lineHeight="md"
+        color="darkest"
+        minHeight={["auto", "72px", "48px"]}
```
iAdramelk, line 55: This should be ok without `minHeight`, or do I miss something?

components/ProductBanner/ProductBanner.tsx
```diff
@@ -0,0 +1,150 @@
...
+import Image from "next/image";
+import { Centrator } from "components/Layout";
+import Flex from "components/Flex";
+import Box from "components/Box";
+
+interface ProductBannerItemProps {
+  title: string;
+  src?: string;
+  width?: number;
+  height?: number;
```
iAdramelk, line 11: I think we can use fixed value for `width` and `height` here, it will not look good if different items have different sizes and it will simplify DX.

components/ProductBanner/ProductBanner.tsx
```diff
@@ -0,0 +1,150 @@
...
+      >
+        <Flex flexDirection={["column", "row"]}>
+          {/* Text Element */}
+          <Flex
+            flexDirection="column"
+            pb={2}
+            pl={3}
+            maxWidth={["auto", "40%"]}
+            order={imgPosition === "right" ? 0 : 1}
+          >
```
iAdramelk, line 98: We have `Heading` component for `title/subtitle`, maybe you can use it here too.

components/ProductBanner/ProductBanner.tsx
```diff
@@ -0,0 +1,150 @@
...
+            <Box
+              color="dark-purple"
+              fontSize="text-md"
+              lineHeight="lg"
+              fontWeight="bold"
+              mb={2}
+            >
+              {subtitle}
+            </Box>
+            <Box
```
iAdramelk, line 108: Please add `as="h2"` to the header if youu need to use custom component in the end and not `Header`.

components/ProductBanner/ProductBanner.tsx
```diff
@@ -0,0 +1,150 @@
...
+              fontSize={["header-3", "header-1"]}
+              lineHeight={["xl", "xxl"]}
```
iAdramelk, line 110: Its my subjective optinion, but I think lineHeight on desktop is too big right now. Please ask Alex about it.

iAdramelk (changes requested):
Very good! Just few small nitpicks from me this time.

components/ProductBanner/ProductBanner.tsx
C-STYR: Agreed, I just wanted to leave the door open if the use case extended beyond icon-like images to something larger.  I have removed it for now.

components/ProductBanner/ProductBanner.tsx
C-STYR: This would require further customization of `Heading`; which is incompatible currently with string[] types for some props and further is too specific on title line heights, fontSizes and fontWeights.  It could be changed but I think not using `Heading` is probably simpler.

components/ProductBanner/ProductBanner.tsx
C-STYR: Done

components/ProductBanner/ProductBanner.tsx
C-STYR: per Alex, we're going to retain "xxl"

PR #366, opened 2021-07-30
Announcement Bar component

C-STYR:
Basic prototyping only; lacks animation functionality currently.


iAdramelk (commented):
Code is ok, but I'm not sure that I understand what this component should be doing. Will it be how logos will change and what will be animated?

Also do we add each logo in separate prop? Maybe add them all as an array with random number of props?

C-STYR:
Screen Shot 2021-08-03 at 12 58 12 PM

Compressed into one component w/ hard-coded logos.

iAdramelk (approved):
Approve with few nitpicks.

PR #367, opened 2021-08-02
Video bar

Alqanar:
added video bar component


components/VideoBar/VideoBar.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const hDisplay = hour > 0 ? `${hour}:` : "";
+
+  return `${hDisplay}${addZero(min)}:${addZero(sec)}`;
+}
+
+export interface VideoBarProps {
+  previewSrc: string;
+  videoLink: string;
+  name: string;
+  videoDurationSec?: number;
```
iAdramelk, line 28:
```suggestion
  thumbnail: string;
  href: string;
  title: string;
  duration?: number;
```

We don't need `video` prefix here, because the is only video here. Also `title` and `href` and the names that we use for such props in other components, so I renamed them, for consistency.

components/VideoBar/VideoBar.tsx
```diff
@@ -0,0 +1,158 @@
...
+  previewSrc,
+  videoLink,
+  name,
+  videoDurationSec,
+}: VideoBarProps) {
+  console.log(previewSrc);
+  const videoDurationMin = secondsToHms(videoDurationSec);
+
+  return (
+    <StyledVideoBar
```
iAdramelk, line 41: Let's by default use standard elements like `Box` and `Flex` and only use styled components if we have nested styles or variations.

This way we can read code in a flow and wouldn't need to scroll back and forth before style definition and component body.

iAdramelk (changes requested):

iAdramelk:
Please check mobile version, button is broken.

image

components/MDX/MDX.tsx
```diff
@@ -52,6 +53,7 @@ export const components = {
   TileImage,
   Figure,
   Notice,
+  VideoBar,
```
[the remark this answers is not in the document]
Alqanar: Yep, I understand, I added it for debug purposes only. Removed

components/VideoBar/VideoBar.stories.tsx
```diff
@@ -0,0 +1,24 @@
+import { Story } from "@storybook/react";
+import MDX from "../MDX";
+import VideoBar, { VideoBarProps } from "./VideoBar";
+import Img from "./assets/video-bar.png";
+
+const StoryComponent: Story<VideoBarProps> = (args) => (
+  <MDX>
+    <VideoBar {...args} />
+  </MDX>
```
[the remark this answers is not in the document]
Alqanar: Removed

components/VideoBar/VideoBar.tsx
```diff
@@ -0,0 +1,158 @@
...
+  videoDurationSec?: number;
+}
+
+export default function VideoBar({
+  previewSrc,
+  videoLink,
+  name,
+  videoDurationSec,
+}: VideoBarProps) {
+  console.log(previewSrc);
```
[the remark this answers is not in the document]
Alqanar: My bad... Removed

components/VideoBar/VideoBar.tsx
Alqanar: fixed

components/VideoBar/VideoBar.tsx
Alqanar: Yes, sure. But I follow the rule - if properties are a lot, I will take them out in styled components. Now I made a mistake in one place, and I fixed it

Alqanar:
> Please check mobile version, button is broken.
> 
> image


Fixed. New version:

Screenshot 2021-08-03 at 01 20 41

components/VideoBar/VideoBar.stories.tsx
```diff
@@ -0,0 +1,20 @@
+import { Story } from "@storybook/react";
+import VideoBar, { VideoBarProps } from "./VideoBar";
+import Img from "./assets/video-bar.png";
```
iAdramelk, line 3: Lets rename folder to `fixture`. `assets` - images that are part of components, `fixtures` – images for tests and storybook.

components/VideoBar/VideoBar.tsx
```diff
@@ -0,0 +1,135 @@
...
+  return (
+    <StyledVideoBar
+      as="a"
+      variant="secondary"
+      href={href}
+      target="_blank"
+      rel="noopener noreferrer"
+    >
+      <StyledImageWrapper>
+        <StyledIcon name="play" size="md" opacity={0.57} />
```
iAdramelk, line 31: IF we want to use `styled` component here, let's move size and opacity there too, to have everything in the same place.

iAdramelk (approved):
Approved with nitpicks.

components/VideoBar/VideoBar.tsx
Alqanar: I moved opacity in the styled components, but "size" is the parameter of the component Icon.

PR #368, opened 2021-08-02
Update README.md

yjperez:
Updating server URL for Docs. Several colleagues ran into issues running the server and the issue turned out to be the wrong URL. As this stumbled several members of the team updating


iAdramelk:
The reason it written this way now is because it's not only docs already. We are migrating old site pages here too. So e. g. `/about` or `/newsletter` are also work. But the fact that we still didn't migrate `/` page is confusing things, its true. So yes, lets change it until we migrate more pages.

PR #369, opened 2021-08-03
video bar integration

Alqanar:
Added the integration possibility for video bar on documentation pages.
To add the video banner on the documentation page, we need to add the key "videoBanner" in the front matter block.
In the key "videoBanner" needs to pass an object with fields: "title", "thumbnail", and "href", also it can be passed field "duration" in the format "mm:ss".


components/DocsPage/DocsPage.tsx
```diff
@@ -77,6 +78,14 @@ const DocsPage = ({
               githubUrl={githubUrl}
               icon={icon}
             />
+            {videoBanner && (
+              <VideoBar
+                thumbnail={videoBanner.thumbnail}
```
iAdramelk, line 83: Nit: You can write `{...videoBar}` here. It will simplify things if we decide to change components API – there will be less places in code to change syntax.

components/VideoBar/VideoBar.tsx
```diff
@@ -56,6 +50,7 @@ const StyledVideoBar = styled(Button)(
     height: "auto",
     fontSize: "text-md",
     fontWeight: "bold",
+    mb: 1,
```
iAdramelk, line 53: Nit: Let's add it in `DocsPage`, we don't want components to have margins by default if we don't use them as part of the content zone. You can make some of the wrapper styling props public like this https://github.com/gravitational/next/blob/main/components/Notice/Notice.tsx#L26

iAdramelk (approved):
Approve with nits.

components/DocsPage/DocsPage.tsx
Alqanar: thx, I fixed it

components/VideoBar/VideoBar.tsx
Alqanar: Yes, you talk about a good thing, but in this case, if margin-bottom cancel, we will have element without box-shadow, and our video banner visually merge with the next block.

Alqanar: Moved the margin-bottom and box-shadow properties to the external properties of the videoBar component

Issue #322, opened 2021-06-21
Video bar


klizhentas:
Create a react plugin that renders video bar:

https://app.avocode.com/revisions/81309351/?design=54379642

Issue #222, opened 2021-04-29
Design narrow play band


klizhentas:
We want to create a youtube component that is a narrow "play" band that we can use on top of the page. The current version takes too much space.

@alexwolfe will provide the design and then we'll need a react component.

alexwolfe:
avocode link: https://app.avocode.com/view/109b3ddb4d6a4f52a92933d3a4cbbd9a/_/inspect

![image](https://user-images.githubusercontent.com/407237/116757147-61883c80-a9c2-11eb-9e19-753e37a16e99.png)

PR #370, opened 2021-08-04
video meta for the video bar

Alqanar:


iAdramelk (approved):
Code is ok, with couple nitpiks, but small mobile version looks not very good with 4 client in  one row. Maybe make it 2 instead of 4?

Also I think line height it title on mobile is too large.

Can you discuss it with @alexwolfe?

utils/youtube-meta.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+const addZero = (num: number) => String(num).padStart(2, `0`);
+
+export async function fetchVideoMeta(id: string): Promise<Meta | FullMeta> {
+  let data: Meta | FullMeta = {
+    href: `${YOUTUBE_URL}?v=${id}`,
+  };
+
+  if (!NEXT_PUBLIC_YOUTUBE_API_KEY && process.env.NODE_ENV === "development") {
```
iAdramelk, line 71: Nit: Let's remove `process.env.NODE_ENV === "development"`. We may want to run local build to test some stuff without the key too.

utils/youtube-meta.ts
```diff
@@ -0,0 +1,105 @@
...
+    data = {
+      href: "#",
+      title: "This is a fake video link for development mode",
+      thumbnail:
+        "https://i.ytimg.com/vi/A93bGm4I2a4/hqdefault.jpg?sqp=-oaymwEcCNACELwBSFXyq4qpAw4IARUAAIhCGAFwAcABBg==&rs=AOn4CLCNpZMhxXTaUdaTDBjm3aFZvjJD3Q",
```
iAdramelk, line 76: Nit: Let's use local placeholder image here. Any existing video on youtube may be deleted at any point of time with its thumbnails.

utils/youtube-meta.ts
```diff
@@ -0,0 +1,105 @@
...
+interface RawItem {
+  contentDetails: RawContentDetails;
+  snippet: RawSnippet;
+}
+
+interface RawVideoMeta {
+  items: RawItem[];
+}
+
+const addZero = (num: number) => String(num).padStart(2, `0`);
```
iAdramelk, line 64: Nit: Let's move it before `changeFormatDuration`.

utils/remark-import-frontmatter.ts
```diff
@@ -20,16 +22,20 @@ interface RemarkImportFrontmatterOptions {
...
     const firstChild = children[0];
-    let config = {};
+    let config: { videoBanner?: string | Meta | FullMeta } = {};
 
     if (firstChild && firstChild.type === "yaml") {
       config = yaml.load(firstChild.value as string) as Record<string, unknown>;
     }
 
-    root.children.unshift(createMdxjsEsmNode(name, config));
+    if ("videoBanner" in config && typeof config.videoBanner === "string") {
```
iAdramelk, line 35: Let's add an option to the plugin that will allow us to disable video fetching and remove `videoBanner` from the config.

This way, if we don't have internet locally or have some problems with the hosting we can temporarily disable banners.

iAdramelk (commented):
Good work! Just a few small nitpics.

iAdramelk (approved):

Alqanar:
https://github.com/gravitational/next/issues/322 https://github.com/gravitational/next/issues/222
Added video bar component.
To use this component, have to pass to Front Matter the "videoBanner" key with the YouTube video identifier in .mdx file. The ID is stored in the URL in the part that follows the “?v=...” characters.

PR #373, opened 2021-08-06
EasyStart Component

C-STYR:
Screen Shot 2021-08-06 at 11 06 21 AM


components/EasyStart/EasyStart.tsx
```diff
@@ -0,0 +1,125 @@
...
+                color="dark-purple"
+                fontSize={["text-md", "text-md", "text-xl"]}
+                lineHeight={["lg", "xxl"]}
+                fontWeight="bold"
+              >
+                Teleport consists of just two binaries.
+              </Box>
+              <StyledOL>
+                <li>
+                  The <StyledSpan>tsh</StyledSpan> client allows users to login
```
iAdramelk, line 61: Two comments here:

1. We should use `code` here, instead of `span`.
2. We have different styles that we use for code blocks on other pages, so we probably should unify them. See STEP 1 on this page for example https://goteleport.com/features/access-requests/ .

components/EasyStart/EasyStart.tsx
```diff
@@ -0,0 +1,125 @@
...
+                  The <StyledSpan>teleport</StyledSpan> agent can be installed
+                  on any server or any Kubernetes cluster with a single command.{" "}
+                </li>
+              </StyledOL>
+              <Button shape="lg" width={["auto", "230px", "230px"]}>
+                Download Teleport
+              </Button>
+            </Flex>
+            {/* Console */}
+            <Flex width={["auto", "50%"]}>
```
iAdramelk, line 74: If we leave this block inside `{children}`, we can render it using code block ``` syntax, it will automatically add syntax highlight on render. See https://github.com/gravitational/next/blame/main/pages/features/access-requests/index.mdx#L55-L70 for an example.

If we want to hardcode it, we should use for `hl.js` classnames and styles from the current `code` component, so if will not be broken in the case we change code theme in the future. See current classes here https://github.com/gravitational/next/blob/main/components/Code/Code.tsx

iAdramelk (changes requested):

components/EasyStart/EasyStart.tsx
C-STYR: Spoke to Alex, we're going to nix the generic black box and move to Terminal component in .mdx file.

components/EasyStart/EasyStart.tsx
C-STYR: done

components/EasyStart/EasyStart.tsx
```diff
@@ -0,0 +1,125 @@
...
+import css from "@styled-system/css";
+import { Centrator } from "components/Layout";
+import Flex from "components/Flex";
+import Box from "components/Box";
+import Button from "components/Button";
+import wavelight from "./assets/wave-light.png";
+
+export const EasyStart = () => {
+  return (
+    <Flex background={`url(${wavelight}) 0 0 no-repeat`}>
```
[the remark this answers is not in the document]
C-STYR: done

components/EasyStart/EasyStart.tsx
```diff
@@ -0,0 +1,125 @@
...
+          <Flex maxWidth={["auto", "75%"]} flexDirection="column">
+            <Box
+              as="h2"
+              fontSize={["header-2", "header-1"]}
+              fontWeight="black"
+              color="black"
+              lineHeight="xxl"
+              mb={[1, 3]}
+            >
+              Easy to get started
```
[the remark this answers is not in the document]
C-STYR: Per Alex, we are going to keep these props hard-coded.  This banner will exist on all four products pages as well as the homepage and in order to enforce consistency props will not be modifiable in .mdx.

PR #374, opened 2021-08-06
MessageBanner component

C-STYR:
Screen Shot 2021-08-06 at 12 47 41 PM


components/MessageBanner/MessageBanner.tsx
```diff
@@ -0,0 +1,65 @@
+import styled from "styled-components";
+import css from "@styled-system/css";
+import Link from "next/link";
+import Flex from "components/Flex";
+import Box from "components/Box";
+import { Centrator } from "components/Layout";
+import doubleWave from "./assets/wave-double.png";
```
iAdramelk, line 7: Can we use `Section` component here instead? Ideally we want to limit the number of background patterns to a limited list of variations instead of adding every possible variation. So I think we should either use exisitng one or add this to the options. Can you please ask @alexwolfe about it?

components/MessageBanner/MessageBanner.tsx
```diff
@@ -0,0 +1,65 @@
...
+      >
+        <Box
+          as="h3"
+          fontSize={["header-2", "header-1"]}
+          fontWeight="black"
+          color="black"
+          lineHeight={["xl", "xxl"]}
+          mb={3}
+        >
+          Our vision for the future
```
iAdramelk, line 33: Shouldn't we place data to the `title`, `children` and `href` props instead? Or are we sure that will not use this component anywhere else? If we don't plan to use it anywhere else maybe its better to replace it with some component that we may use in different places?

I think or goal with old pages migration is not to port existing pages as is, but to unify their designs and create constructor of reusable components too. CC @alexwolfe

components/MessageBanner/MessageBanner.tsx
```diff
@@ -0,0 +1,65 @@
...
+        </Box>
+        <Box
+          fontSize={["text-lg", "header-4"]}
+          lineHeight="lg"
+          color="darkest"
+          mb={3}
+          maxWidth={["90%", "70%"]}
```
iAdramelk, line 40: Let's make it `maxWidth={["auto", "70%"]}`.

It's too big on mobile right now. See:

image

C-STYR: per Alex, leave as it is...

components/MessageBanner/MessageBanner.tsx
C-STYR: agreed

components/MessageBanner/MessageBanner.tsx
C-STYR: Sure, np.  I have now reorganized `Section` and alphabetized.

PR #383, opened 2021-08-12
Command snippet

### LINKED ISSUE #96: When selecting code ignore `$` sign.
We currently provide code snippets that are copied and pasted by people. It's a small annoyance but we should remove the `$` from the copying and paste. A ton of dev tools do this, but heroku is a good example. 

 https://devcenter.heroku.com/articles/heroku-cli

### Before
image

### UX After change 
image

### LINKED ISSUE #166: Code snippet

### LINKED ISSUE #275: Code snippet component

Alqanar:
Fix #96
Fix #166 
Fix #275 

This plugin intended for transform code blocks which should be with ability copy command.

At the start of the line in MDX, we write ` ```code  `. After this expression, all our code will be transformed with the help of the plugin.
To highlight the command, we have to use the symbol `$`. It will be mean that this line is a command and will be transformed to a command component with the ability to copy its content by button.
After `$`, we can type space or some spaces, or we can don't type space. The plugin will delete the `$` and spaces before starting the line and change on `$` with one space.
If line started with `$` and at the end of line `\`, the plugin will recognize that it is a multiline command and process it.
All lines without `$` will look the same as comments.

The screenshot shows a code example, and the video shows what it looks like and how it works

Screenshot 2021-08-17 at 20 40 41

https://user-images.githubusercontent.com/41822696/129848894-ce377e63-7e6b-4f73-9abc-370eb117f5a8.mov


iAdramelk (approved):
Approve with couple of nits

PR #392, opened 2021-08-20
3rd level nav view

### LINKED ISSUE #390: Finish 3rd level nav design
Design as currently implemented has issues:

3rd-level

With a 3rd level present, the sections all merge together and hard to read.

The original design addresses this issue:

3rd-level-required

https://app.avocode.com/revisions/81988089/?design=51942844

Implement the original design

Alqanar:
Fix #390 

Submenu items are hidden if the item page is not open, the menu item is still active if its sub-items are selected, added arrows to the menu items on the second level, where there are additional items.


klizhentas:
@alexwolfe preview here:

https://teleport-docs-next-git-fix-menu-levels-gravitational.vercel.app/docs/setup/admin/

@alexwolfe @Alqanar should it collapse once you click back? Otherwise looks OK to me, may be another small nit is that the font on the 3rd level is getting quite small

alexwolfe:
Here are some updates that need to be made to the current navigation and thoughts behind it: 
First our navigation actually works differently than first designed which is the reason for the changes. 

1. The top level buttons are expanding and collapsing in their functionality. This was not present in the original design. As such, we need to indicate that can be open and closed with the arrow icon, currently they have no indicator. 
2. The second level navigation should not have the arrows since they cannot be expanded or collapsed. The link itself is an actual page not a toggle button that takes the user to a unique url. Simply removing the arrow will fix this issue. Once a user clicks on this section they will see the nested content but clicking on the parent simply goes to the parents page
3. When an item is selected the background must change. Changing the background helps to highlight what section is selected and also makes the content easier to read. If the third tier item is selected the background for the third tier content should be selected while the text for the second tier simply changes to purple. 

image

components/DocsPage/Navigation.tsx
```diff
@@ -70,22 +79,30 @@ const DocNavigationCategory = ({
   );
 
   return (
-    <Box as="nav" key={title}>
+    <Fragment>
```
iAdramelk, line 82: NIT: Let's use `<>` syntax here. We already using it in this file above.

components/DocsPage/Navigation.tsx
```diff
@@ -194,20 +217,46 @@ const CategoryButton = styled(HeadlessButton)(
...
-    lineHeight: "lg",
-    color: active ? "dark-purple" : "gray",
-    fontWeight: active ? "bold" : "regular",
-    textDecoration: "none",
-    "&:focus, &:hover, &:active": {
-      outline: "none",
-      bg: "white",
+    display: "none",
+
+    "& a": {
```
iAdramelk, line 247: Can we use something like `['& ${Component}']` here?

components/DocsPage/Navigation.tsx
```diff
@@ -150,23 +167,28 @@ const DocNavigation = ({ data, section }: DocNavigationProps) => {
...
-            opened={index === openedId}
-            onToggleOpened={setOpenedId}
-            onClick={toggleMenu}
-            {...props}
-          />
-        ))}
+        <Box as="ul" listStyle="none">
+          {data.map((props, index) => (
+            <Box as="li" key={index}>
+              <DocNavigationCategory
```
iAdramelk, line 181: I think we should make this part recursive, there is nothing that prevents us from having 4th level of menu in the config. But let's style all levels below the 3rd the same way.

iAdramelk (approved):

PR #399, opened 2021-08-24
Code plugin improvements

### LINKED ISSUE #393: Code component improvements
Let's further improve the component. Right now, when you click `copy`  on the right side, it copies all the content verbatim, but removes the dollar sign. 

Here is what I'd like to change:

* [x] Keep the behavior and design of `$`.
* [x] Remove the `#` from lines, including optional space and mark those as comments
* [x] When someone clicks copy on the right side, skip the lines starting with `#` from the copy output.
* [x] If someone adds `##` in front of the line, remove the `#` and keep the other one, so people can show real comments. Add the comment in the copy output contents.

```code
$ tctl status

# These lines will be omitted from copy including the optional space after `#`
# Cluster  acme.example.com
# Version  (=teleport.version=)
# CA pin   sha256:sha-hash-here

## This will be rendered as a comment starting from one # and will be included in the copy
```

* [x] Support bash [heredoc](https://tldp.org/LDP/abs/html/here-docs.html) format:  `<<EOF`

This should render properly and the entire line with heredoc should be one command:

```code
# Copy and Paste the below and run on the Teleport Auth server.
$ cat > api-role.yaml <<EOF
kind: role
metadata:
  name: api-role
spec:
  allow:
    rules:
      - resources: ['role']
        verbs: ['read']
  deny:
    node_labels:
      '*': '*'
version: v3
EOF
# Create role
$ tctl create -f api-role.yaml
# Add user and login via web proxy
$ tctl users add api-user --roles=api-role
```
Detect the format `<<DELIMITER`:

* [x] Support extended bash format that prints hostname: 

```code
# hostname $ should be ignored and removed alongside `$` sign
hostname $ teleport status
# same behavior
$ teleport status
```

Alqanar:
Fix #393 

**Support extended bash format that prints hostname**

To use it, use the sign > at the start of the string and the sign $ where you want the command to begin
For example:
1. In mdx file we will write this - pic_1
2. It will be render to this command - pic_2
3. Only the area highlighted in red will be copied - pic_2

pic_1
pic_2


utils/fixtures/result/code-snippet-multiline.mdx
```diff
@@ -0,0 +1,15 @@
+## Header
+
+<Snippet>
+  <Command>
+    <span data-content="$ ">curl \\</span>
```
iAdramelk, line 5: Looks like we need to remove space and the end here.

iAdramelk (changes requested):
Good job, but let's add more tests to make use errors didn't break builds!

iAdramelk (approved):

PR #410, opened 2021-09-01
Marketo recaptcha integration

iAdramelk:
This PR adds recaptcha integration:

- Load recaptcha code.
- Add own recaptcha instance and callbacks to each form.
- Add recaptcha token to requests.
- Validate token on the server side before submitting data to Marketo.


Alqanar (approved):
LGTM

components/MarketoForm/helpers.ts
```diff
@@ -70,44 +71,18 @@ export const useValidate = ({
...
+  fields: Record<string, string>,
+  token: string
 ) => {
   try {
     const response = await fetch(`/api/form/${id}/`, {
       method: "POST",
       body: JSON.stringify(fields),
       headers: {
         "Content-Type": "application/json",
+        "g-recaptcha-response": token,
```
alex-kovoy, line 85: if we are sending some data to our backend, i would validate the fields (like maxLengths) on the server and also make sure that there is a maximum request size is set (or handled). It might be handled already but still worse mentioning it.

alex-kovoy (approved):

components/MarketoForm/helpers.ts
iAdramelk: Client validation is based on marketo fields config that we receive from their API. Our server just proxies request to the `marketo` API after we check for recaptcha token. And marketo backed have these validations built in. So even if we send fields with too long length, `marketo` API will just return up an error and we display them.

alex-kovoy: Correct but our sever still needs to process a request. After a quick check I found this:

> By default, express. json() limits the request body to 100kb. If the request body is any larger, Express will throw an HTTP 413 "Payload Too Large" error

So we should be fine then if we use this middlewere.

C-STYR (approved):

PR #412, opened 2021-09-01
Creates product pages: Server, Kubernetes, Database Access

C-STYR:


pages/products/database-access/index.mdx
```diff
@@ -0,0 +1,179 @@
...
+  MessageBanner,
+  ProductBanner,
+  Section,
+  TerminalAnimation,
+  Terminal,
+  TryTeleport,
+  VideoBanner,
+  WorksWith,
+} from "components";
+import GridDisplay, { GridTile } from "components/GridDisplay";
```
iAdramelk, line 19: Shouldn't it be part of common export?

pages/products/database-access/index.mdx
```diff
@@ -0,0 +1,179 @@
...
+    }
+  />
+  <AnnouncementBar />
+</Section>
+<Section bg="wavelight">
+  <ProductBanner
+    imgPosition="right"
+    title="Secure your databases & meet compliance requirements"
+    subtitle="For DevSecOps"
+    src="../assets/server-access.png"
```
iAdramelk, line 39: Can we move assets to the page's subfolder instead of common folder? If we do it, it will be easier to remove unused assets then we delete/update page.

pages/products/database-access/index.mdx
```diff
@@ -0,0 +1,179 @@
...
+</MessageBanner>
+<TryTeleport/>
+<GridDisplay
+  centralHeading={true}
+  title="Learn more about Teleport Database Access"
+  subtitle="Explore resources"
+  titleFontSize="header-1"
+  titleFontWeight="black"
+  >
+  <GridTile
```
iAdramelk, line 135: Is this one-time component? Maybe move reusable component from it?

iAdramelk (commented):
I didn't check small fixes inside components, because I assume you approved them with Alex.

I have some nitpicks about page files, wrote them in the first file only, but them apply for all 3. Please check.

pages/products/database-access/index.mdx
C-STYR: No, this design pattern will see further use.  GridDisplay, GridCard and GridTile can be restructured if necessary.

pages/products/database-access/index.mdx
C-STYR: No, the files share 90% of the images, there is no reason to create copies for each subfolder.

iAdramelk (approved):

PR #416, opened 2021-09-07
added rules for the mdx linter

Alqanar:
When the linter rules are checked, we will see errors in the command component if there are any.

Screenshot 2021-09-07 at 20 02 18


iAdramelk (changes requested):
Let's add linter to the tests too. See remark-includes for an example.

iAdramelk (approved):

PR #433, opened 2021-09-16
Update Mac signing cert information

wadells:
These were rotated in:

  https://github.com/gravitational/teleport/pull/8230

However the majority of published binaries are still signed with the old
certificate, so we need to keep documentation about it around for the time being.

## Testing Done
Eyeballed https://teleport-docs-next-3wn11jkhv-gravitational.vercel.app/security/

image


webvictim:
Just FYI this is what the actual CLI output looks like on a Mac:

```
~ » sw_vers
ProductName:	macOS
ProductVersion:	11.6
BuildVersion:	20G165

# old signature on 7.0.0 artifact
~ » pkgutil --check-signature teleport-7.0.0.pkg
Package "teleport-7.0.0.pkg":
   Status: signed by a developer certificate issued by Apple for distribution
   Signed with a trusted timestamp on: 2021-08-10 20:49:27 +0000
   Certificate Chain:
    1. Developer ID Installer: Gravitational Inc. (QH8AA5B8UP)
       Expires: 2024-11-14 22:40:03 +0000
       SHA256 Fingerprint:
           78 05 14 69 20 59 21 D1 EE 96 42 01 5A 28 35 FB E1 D4 38 5E 2A 23
           5D 62 73 A4 D1 27 8A 33 BA 34
       ------------------------------------------------------------------------
    2. Developer ID Certification Authority
       Expires: 2027-02-01 22:12:15 +0000
       SHA256 Fingerprint:
           7A FC 9D 01 A6 2F 03 A2 DE 96 37 93 6D 4A FE 68 09 0D 2D E1 8D 03
           F2 9C 88 CF B0 B1 BA 63 58 7F
       ------------------------------------------------------------------------
    3. Apple Root CA
       Expires: 2035-02-09 21:40:36 +0000
       SHA256 Fingerprint:
           B0 B1 73 0E CB C7 FF 45 05 14 2C 49 F1 29 5E 6E DA 6B CA ED 7E 2C
           68 C5 BE 91 B5 A1 10 01 F0 24

# new signature on 7.1.2 artifact
~ » pkgutil --check-signature teleport-7.1.2.pkg
Package "teleport-7.1.2.pkg":
   Status: signed by a developer certificate issued by Apple for distribution
   Signed with a trusted timestamp on: 2021-09-15 00:49:03 +0000
   Certificate Chain:
    1. Developer ID Installer: Gravitational Inc. (QH8AA5B8UP)
       Expires: 2026-07-27 18:27:29 +0000
       SHA256 Fingerprint:
           78 2F E1 18 5F A1 AD 68 AD 25 0B A9 4D 21 DC BB 0D 8E 47 C6 E4 1D
           FE FB AB 05 41 33 4C 33 1D 43
       ------------------------------------------------------------------------
    2. Developer ID Certification Authority
       Expires: 2027-02-01 22:12:15 +0000
       SHA256 Fingerprint:
           7A FC 9D 01 A6 2F 03 A2 DE 96 37 93 6D 4A FE 68 09 0D 2D E1 8D 03
           F2 9C 88 CF B0 B1 BA 63 58 7F
       ------------------------------------------------------------------------
    3. Apple Root CA
       Expires: 2035-02-09 21:40:36 +0000
       SHA256 Fingerprint:
           B0 B1 73 0E CB C7 FF 45 05 14 2C 49 F1 29 5E 6E DA 6B CA ED 7E 2C
           68 C5 BE 91 B5 A1 10 01 F0 24
```

Couple of thoughts:

1) We should probably publish SHA256 hashes rather than SHA1, if that's what MacOS is displaying by default? `pkgutil` doesn't appear to have a CLI flag to print SHA1 instead of SHA256 hashes.
2) We should use 7.1.2 in the example to indicate that's the first version where we'd expect to see the new signature hash. I know it's only an example of usage, but if people were to run the command and get different output they might be alarmed.

wadells:
>     1. We should probably publish SHA256 hashes rather than SHA1, if that's what MacOS is displaying by default? `pkgutil` doesn't appear to have a CLI flag to print SHA1 instead of SHA256 hashes.

Good catch -- this looks like it changed sometime.  Different OSX versions have different behavior :frowning_face: [According to strangers on the internet](https://security.stackexchange.com/a/226416), this happened around 10.14 -> 10.15.

I'll publish both figureprints, and update the command output to show the new SHA256, as that is way more future proof.

>     2. We should use 7.1.2 in the example to indicate that's the first version where we'd expect to see the new signature hash. I know it's only an example of usage, but if people were to run the command and get different output they might be alarmed.

Excellent point as well.

C-STYR (approved):

PR #434, opened 2021-09-16
Update fingerprint styling to monospace

wadells:
We want these to have a fixed width font, because they're code, not prose.

## Testing Done
Checked it out at https://teleport-docs-next-o67fyk0yg-gravitational.vercel.app/security/

image


components/PubCertsKeys/PubCertsKeys.tsx
```diff
@@ -279,6 +279,8 @@ Extensions    : {System.Security.Cryptography.Oid, System.Security.Cryptography.
 const CodeSpan = styled("span")(
   css({
     backgroundColor: "lighter-gray",
+    fontFamily: "monospace",
+    px: 1,
```
wadells, line 283: `px: 1` seems a bit wide, but I have no idea what a better incantation would be.  No left/right padding at all looked worse. cc @alexwolfe

C-STYR (approved):
I'm fine with these changes; we can always revise later when time is more ample.

PR #436, opened 2021-09-16
Disable marketo if we don't have ENV variables

iAdramelk:


C-STYR:
Looks good once the linting issue is fixed...

iAdramelk:
> Looks good once the linting issue is fixed...

@C-STYR missed a file here. ^^" Pushed it now.

C-STYR (approved):

PR #437, opened 2021-09-17
Application Access page and new Animation

C-STYR:
- adds Application Access page
- adds new animation 
- restructures existing animation files and utilities

Reviewers: the animation is complete as designed, however I think it lacks a clear transition between the last two screens; perhaps some visual clicking element on the Jenkins logo in the penultimate screen?


components/AnimationHeader/AnimationHeader.tsx
```diff
@@ -73,7 +76,8 @@ export const AnimationHeader = ({
         alignItems="center"
         maxWidth="100%"
       >
-        <TerminalAnimation text={terminalText} />
+        {animation === "terminal" && <TerminalAnimation text={terminalText} />}
+        {animation === "application" && <ApplicationAnimation />}
```
iAdramelk, line 80: Do we need to store all animations in one component? Right now all animations are added to the bundle and are downloaded for each page, even if they are not used there. Maybe separate them?

components/AnimationUtilities/ApplicationBrowserAnimation.tsx
```diff
@@ -0,0 +1,260 @@
...
+  StyledVerifyButton,
+  StyledCenterBox,
+  StyledImageContainer,
+  BrowserClose,
+  BrowserMin,
+  BrowserMax,
+  CloseTabButton,
+  AddTabButton,
+} from "./AnimationUtilities";
+import { check, biometric, password } from "./assets";
```
iAdramelk, line 21: Nit: I may be mistaken, but I think we should probably use larger blocks here. Like `Browser` as wrapper with all buttons and UI, instead of using them as parts. It will make creating new animations easier.

components/AnimationUtilities/ApplicationBrowserAnimation.tsx
```diff
@@ -0,0 +1,260 @@
...
+          clearInterval(kennyLogin);
+        }
+      }, 60);
+    }, 1500);
+  };
+
+  useEffect(() => {
+    enterPassword();
+  }, []);
+
```
iAdramelk, line 47: If we unmount this component before animation is finished, calling `setInputState` may start throwing error. We probably want to stop animation on component unmount because of them.

components/AnimationUtilities/AnimationUtilities.tsx
```diff
@@ -152,3 +155,102 @@ export const AnimatedTerminal = styled(Terminal)(
...
 );
+
+//input field and verify button for simulated sign-in screen
+export const StyledInput = styled("input")(
+  css({
+    border: "1px solid #c5c5c5",
+    borderRadius: "4px",
+    boxSizing: "border-box",
+    color: "gray",
+    fontSize: "18px !important",
```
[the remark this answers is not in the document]
C-STYR: Correct! At some point during dev it was being superseded but is now not necessary. Good catch.

components/AnimationHeader/AnimationHeader.tsx
C-STYR: > Hi @C-STYR , I checked PR, but not in a deep, its too big to load it inside the head all in once.
> 
> If we are time constrained now, we can release it as is, but if we have time for some refactoring I would try to simplify it. For example I think we probably better replace Jenkins screen (and some others) with simple `svg` instead of re-implementing it in browser.
> 
> In the current form we will spend couple of weeks on each new animation and they will be hard to maintain. We can Also discuss with Alex how to make animation steps more unified and reusable to streamline the process of their creation.

The animation was originally built with `svgs` for the last two screens; however they were insufficient for our needs at various screen sizes given the limitations of our breakpoints (and for other reasons); I agree that re-implementation of the browser was a more complex process but it has yielded the desired result.  

We are _always_ time constrained, and this animation was no different.  Alex and I have agreed that refactoring this animation (for example the 'click' simulation I have suggested in my first comment) is a low priority presently given other issues in the queue.  

Styled System (with Styled Components) is a non-optimal platform for animations. Reusability is the first casualty as creating animations contingent on state or props is much more complicated than in a traditional vanilla CSS approach.  However I am glad to listen to specific suggestions or comments.  I think an in-depth code review may be necessary when you have time. 

> Do we need to store all animations in one component? Right now all animations are added to the bundle and are downloaded for each page, even if they are not used there. Maybe separate them?

You're probably right, what did you have in mind here?

components/AnimationUtilities/ApplicationBrowserAnimation.tsx
C-STYR: I agree in theory and have done some limited refactoring here with new `BrowserTop` component. However the outer `wrapper` components are animated and require different handling.

components/AnimationUtilities/ApplicationBrowserAnimation.tsx
iAdramelk: My understanding is that setTimeout/setInterval is not removed from the memory then you remove the component from the DOM. So it will be called after component is unmounted and call `setInputState` when the component is no longer exist. Calling `setInputState` in this situation may cause error (or may not, react can have some failsafe built-in for this case).

components/AnimationHeader/AnimationHeader.tsx
iAdramelk: > You're probably right, what did you have in mind here?

I'd just create two separate components for `<ApplicationAnimation />` and `<TerminalAnimation />` and import them directly.

components/AnimationUtilities/ApplicationBrowserAnimation.tsx
C-STYR: Ah, I see what you mean.  Yes, the code executed `setTimeout/Interval` is executed in a different execution context independent of the function/component that calls them.  However, the mechanism for removing the element (the button) cannot be clicked until after the code executed by both `setTimeout` and `setInterval` called by `enterPassword` _have completed_.  Even knowing that the button is there but merely invisible, I was unable after dozens of tries to produce an error by clicking it. If you can produce an error please let me know.

PR #458, opened 2021-09-22
added new page case-study/ecmwf

Alqanar:


components/Section/Section.tsx
```diff
@@ -84,9 +84,9 @@ export interface SectionProps {
   children: React.ReactNode;
 }
 
-export const Section = ({ bg, children }: SectionProps) => {
+export const Section = ({ bg, children, ...props }: SectionProps) => {
```
iAdramelk, line 87:
```suggestion
export const Section = ({ bg, children, ...props }: SectionProps & BoxProps) => {
```

But it first import `BoxProps` and the top of the file.

pages/case-study/ecmwf/index.mdx
```diff
@@ -0,0 +1,231 @@
...
+<Section
+  bg="wavelight"
+  color="darkest"
+  backgroundPosition="top left"
+  backgroundSize="auto"
+>
+  <Centrator mt={[7, 11]} mb={[0, 8]} flexDirection="column">
+    The ECMWF needs to safely and securely grant remote shell access to a supercomputer while knowing little about the end users. So when they modernized, European Centre for Medium-Range Weather Forecasting (ECMWF) chose Teleport. On one hand, the ECMWF was founded to provide mission-critical weather services to its members across Europe. On the other, they had to protect computing clusters that forecast models from unwanted access. The catch? A healthy portion of their ecosystem consisted of third party users connecting remotely from a Member State. In other words, the ECMWF had to open their facilities to researchers they did not have explicit contracts with and cannot know in advance. With Teleport, they have enough control to feel comfortable exposing endpoints to anyone its Member States had authorized.
+
+    <Quote mt={6}>
```
iAdramelk, line 50: We want to make a constructor for content editors to use themselves here, so in the perfect world we want then to just write markdown here instead of using any of custom react element.

So instead of styling elements by hand we should use `MDX` provider to set element styles for things like `###` or `[link](href)`. If the current styles that se use on the `SitePage` are too different from the one we want we should either discuss it with @alexwolfe to check if we can unify them, or if not, we should create separate layout here with new values to MDX provider.

Lets discuss how to better do it then you have a time.

pages/case-study/ecmwf/index.mdx
```diff
@@ -0,0 +1,162 @@
...
+    ![Teleport Web Console](./assets/terminal-web-console.png)
+  </Figure>
+
+  Even for those experienced with Linux, the Web UI has been a welcome addition to the ECMWF. A widely adopted OSS data science tool, JupyterHub, exists in-browser. By being able to hop between services by just switching tabs makes for a much more consistent experience for many data scientists.
+
+  ## Looking Ahead
+  In 1975, the ECMWF embarked on a grand expedition; to foretell the weather with great accuracy. Their move to Italy marked the latest milestone. Committing to lofty goals for 2025, the ECMWF had to put more on the line. More compute. More people. More predictions. More access. Particularly, shell access to the most powerful computers in the world across an ever-changing base of users.
+
+  And yet still, despite an accelerated implementation, the ECMWF was met with overwhelmingly positive feedback.
+  <Flex flexDirection="column" alignItems="center" width="100%" my={3}>
```
iAdramelk, line 149: Can we replace this block with normal image + link underneeth?

pages/case-study/ecmwf/index.mdx
```diff
@@ -0,0 +1,162 @@
...
+  ```
+
+  By connecting the user login with an identity provider, identity information gets propagated to restrict activity, enforce logins, and control sessions.
+
+  ## The Web UI
+  In addition to a local agent, Teleport provides a fully featured web-based SSH console that can be used in-browser. This unlocked an entirely new cast of users that existed solely over the web. The demographic included people that had only used Windows machines and might still be using one. For example some security policies prevent users from downloading any agents on local devices, either for OpenSSH or Teleport. Others just found using the CLI daunting. For all these people, having a web client gave them a channel they previously had not.
+  <Figure
+    xMargin={0}
+    width={600}
+    height={164}
```
iAdramelk, line 136: We can add `rehypeImages` to site mdx config, it will automatically add width and height for each image, so we don't need to do it by hand.

pages/case-study/ecmwf/index.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+  ## What is ecAccess?
+  The two computational behemoths housed by the ECMWF include a high-performance computing Linux cluster and two <Link scheme="site" href="https://en.wikipedia.org/wiki/Cray_XC40">Cray XC40s</Link> that constitute its supercomputer. Capable of up to hundreds of teraflops in processing power, these facilities are used by researchers to crunch data and build machine learning models all through remote shell access via SSH. Such low-level access opened an avenue for hackers to siphon supercomputing resources, with a recent example being a coordinated attempt to use other European supercomputers to <Link scheme="site" href="https://www.zdnet.com/article/supercomputers-hacked-across-europe-to-mine-cryptocurrency/">mine cryptocurrencies</Link>. For the ECMWF, the stakes are higher as their systems directly support emergency services. Days or even weeks of downtime is not tolerable.
+  <Figure
+    xMargin={0}
```
iAdramelk, line 40: We always write `xMargin={0}` here. Can we either move it to `MDXProvider` or ask Alex if we can use standard margin here?

pages/case-study/ecmwf/index.mdx
```diff
@@ -0,0 +1,162 @@
...
+  </Figure>
+
+  ## Servicing the Ecosystem
+  The ECMWF’s user base includes two very different groups. Those with the most privileges included staff, such as analysts and sysadmins. The bulk of users were granted access by the ECMWF’s Member States; a revolving door of scientists and researchers from universities, public utilities, and federal agencies all across Europe. This meant the ECMWF had little control over where a client machine was located, what software it ran, and who sat on the other end.
+
+  Gating access with these limitations proved to be quite difficult. Paradoxically, upgrading security measures had the potential to weaken the ECMWF. For example, SSH keys are superior to password-based authentication for a  <Link scheme="site" href="https://lwn.net/Articles/369703/">variety of reasons</Link>, but issuing SSH keys was never a viable option for the ECMWF. Keys would have to be distributed to a changing user base with little control over what happens to them. Despite being a better method of primary authentication, leaving keys across thousands of local machines meant researchers could access supercomputers even after they’ve left the organization.
```
iAdramelk, line 51: Lets use markdown link instead of component Link in all places we use link.

pages/case-study/ecmwf/index.mdx
```diff
@@ -0,0 +1,162 @@
...
+  Flex,
+  Feedback,
+  Quote,
+} from "components";
+import NextImage from "next/image";
+
+<CaseStudy subtitle="Customer Case Study" title={meta.articleTitle} companyLogo={meta.companyLogo} poster="./assets/ecmwf.png">
+  The ECMWF needs to safely and securely grant remote shell access to a supercomputer while knowing little about the end users. So when they modernized, European Centre for Medium-Range Weather Forecasting (ECMWF) chose Teleport. On one hand, the ECMWF was founded to provide mission-critical weather services to its members across Europe. On the other, they had to protect computing clusters that forecast models from unwanted access. The catch? A healthy portion of their ecosystem consisted of third party users connecting remotely from a Member State. In other words, the ECMWF had to open their facilities to researchers they did not have explicit contracts with and cannot know in advance. With Teleport, they have enough control to feel comfortable exposing endpoints to anyone its Member States had authorized.
+
+  <Quote mt={6} title="About" linkSrc="https://www.ecmwf.int/" linkTextTitle="ECMWF">
```
iAdramelk, line 28: Can we hide `mt` in some way? Either move it to component, or to MDXProvider.

components/Qoute/Quote.tsx
```diff
@@ -0,0 +1,66 @@
...
+            {linkTextTitle}
+          </Link>
+        </Box>
+      )}
+      {children}
+    </StyledQoute>
+  );
+}
+
+const StyledQoute = styled("blockquote")(
```
iAdramelk, line 40: Do we have separate mobile styles for it?

components/MDX/MDX.tsx
```diff
@@ -45,6 +46,7 @@ export const components = {
   td: TD,
   video: Video,
   Admonition,
+  CaseStudy,
```
iAdramelk, line 49: Nit: I'd use direct import for it for now, so we do not forget to remove it after we switch to layout plugin.

iAdramelk (approved):

PR #460, opened 2021-09-24
fixed the search filtering

Alqanar:


components/Search/Search.tsx
```diff
@@ -29,10 +29,11 @@ const Search = ({
           hitsPerPage: ALGOLIA_HITS,
         },
         transformData: function (hits) {
+          console.log(hits);
```
klizhentas, line 32: Let's remove this log entry and merge

PR #466, opened 2021-09-28
Add custom layout support for mdx pages

iAdramelk:
Until now we used the same wrapper for all site pages `SitePage` and it was hardcoded to the `mdx-loader`. But as we are now started working on the new page types like `CaseStudies` it is not very convenient.

This PR adds support for switching layouts using `layout` field in the frontmatter.

```
---
layout: default
---
```

Rest of the frontmatter fields will be added to the components as `meta` prop.

`default` is now equal to `components/SitePage` and is used if `layout` field is not specified. To add new layouts, add them to the `utils/mdx-config-site.ts` file:

```
{
        layouts: {
          default: "components/SitePage",
          newLayout: "pathToComponent"
        },
},
```


utils/remark-layout.ts
```diff
@@ -0,0 +1,47 @@
...
+    const children = root.children as MdxastNode[];
+
+    const firstChild = children[0];
+    let frontmatter: Meta = {};
+
+    if (firstChild && firstChild.type === "yaml") {
+      frontmatter = yaml.load(firstChild.value as string) as Meta;
+    }
+
+    const { layout = "default", ...meta } = frontmatter;
```
iAdramelk, line 41: I assume here that `default` layout is always set, its not good for general purpose plugin, but should work in our case.

components/SitePage/SitePage.tsx
```diff
@@ -19,7 +19,7 @@ interface Props {
     description?: string;
     hideWave?: boolean;
     shortFooter?: boolean;
-    layout?: HeaderMode;
+    headerMode?: HeaderMode;
```
iAdramelk, line 22: Renamed it from `layout` for now. In the future PR I plan to split `SitePage` to three separate components based on `HeaderMode` value to reduce bundle size.

C-STYR (approved):

Alqanar (approved):
LGTM

PR #471, opened 2021-09-30
Fix #454: Add terms pages

### LINKED ISSUE #454: Terms of Service and Privacy Policy
These pages should be be easy to update. The prior tos and privacy pages should be excluded from the site map. 

### Pages
- [x] https://goteleport.com/tos/
- [x] https://goteleport.com/privacy/

We also need to host the prior privacy policies and terms of services

- [x] https://goteleport.com/previous-terms/tos-11-09-2020
- [x] https://goteleport.com/previous-terms/privacy-05-05-2021

iAdramelk:
Added default text layout for terms pages and migrated pages from the old engine to the new one. There are couple minor updates to the base components that I decided to include here too, because they are reeeally minor.

Page urls on staging:

- https://teleport-docs-next-6t12w4hzl-gravitational.vercel.app/tos/
- https://teleport-docs-next-6t12w4hzl-gravitational.vercel.app/privacy/
- https://teleport-docs-next-6t12w4hzl-gravitational.vercel.app/previous-terms/tos-11-09-2020/
- https://teleport-docs-next-6t12w4hzl-gravitational.vercel.app/previous-terms/privacy-05-05-2021/

P. S. Most of the code in the PR is terms content, so don't be afraid of the diff size.


layouts/ContentPage/ContentPage.tsx
```diff
@@ -0,0 +1,91 @@
...
+import SectionHeader from "components/SectionHeader";
+import Head from "components/Head";
+import Layout, { Centrator } from "components/Layout";
+import Footer from "components/Footer";
+import MDX from "components/MDX";
+import Code from "components/Code";
+import Box, { BoxProps } from "components/Box";
+import BaseLink, { LinkProps } from "components/Link";
+
+const components = {
```
iAdramelk, line 11: We starting reusing this style of components a lot. We need to move them to separate place and add default styling for all markdown elements here.

utils/mdx-config-site.ts
```diff
@@ -19,6 +20,7 @@ const config: MdxConfig = {
       {
         layouts: {
           default: "components/SitePage",
+          content: "layouts/ContentPage",
```
iAdramelk, line 23: I plan to move all layout components to separate folder, but left `SitePage` for now there it is, to prevent merge conflicts.

utils/paths.ts
```diff
@@ -11,6 +11,7 @@ const extensions = ["md", "mdx", "ts", "tsx", "js", "jsx"];
 const pagesRoot = resolve("pages");
 const nextPages = [
   new RegExp(`^${pagesRoot}/api/.*$`),
+  new RegExp(`^${pagesRoot}/previous-terms/.*$`),
```
iAdramelk, line 14: We need a better way to handle removing pages from sitemap that this. Old site had ability to set `noindex: true` field in the frontmatter. Maybe implement it too?

pages/tos.mdx
```diff
@@ -0,0 +1,209 @@
...
+
+## Copyright Complaints and Copyright Agent
+
+Gravitational respects the intellectual property of others, and expects users to do the same. If you believe, in good faith, that any materials provided on or in connection with the Services infringe upon your copyright or other intellectual property right, please send the following information to Gravitational’s Copyright Agent at:
+
+```
```
C-STYR, line 124: I think you should specify a neutrally-highlighted language; this code block and the one at the bottom have some random highlights; 
UPDATE: this applies to multiple .mdx files in this PR

pages/privacy.mdx
```diff
@@ -0,0 +1,174 @@
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           | Managing Settings                                                                                                                                                                                                                                                                                                                 |
```
C-STYR, line 64: Table either needs styling or needs styling correction; it lacks similarity to the original.
UPDATE: this applies to multiple .mdx files in this PR

layouts/ContentPage/ContentPage.tsx
C-STYR: I agree! Let's chat about it on Monday

C-STYR (changes requested):

pages/privacy.mdx
iAdramelk: Can you elaborate? Table now uses the same styles as the one in docs https://goteleport.com/docs/adopters/ . They are different from the old ones but I don't think that its a good idea to have few different types of table styles across the site in different places and we spent some time redesigning and updating tables for the docs. I think our goal here is not 1-on-1 visual parity with the original, but also updating styles in the places where we already have updated components.

C-STYR: We will have at least two styles then: the original ones here in /tos and /privacy are the same as those in the How It Works section (see the last three pages where the compliance tables are) and potentially elsewhere.  I have already built a Styled Component to replicate those styles for the HIW page transfer. @alexwolfe perhaps you should make a call here on which table style you want to use going forward, or multiple styles? I personally don't mind either way.

C-STYR: Example of Compliance Table: https://goteleport.com/teleport/how-it-works/soc2-ssh-kubernetes/

alexwolfe:
Hey Team, lets use the original table style for these tables (purple header). We will be updating the Docs table eventually as well (it will not be the same as the marketing site) to give it a slightly better visual design. Thanks, let me know if you have any questions!

C-STYR (approved):

C-STYR:
Approved pending table styling re Alex's comments...

components/SectionHeader/SectionHeader.tsx
```diff
@@ -16,7 +16,7 @@ export interface SectionHeaderProps {
   mode?: string;
   subtitle?: string;
   title: string;
-  children: React.ReactNode;
+  children?: React.ReactNode;
```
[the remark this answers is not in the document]
iAdramelk: Because I didn't know about `PropsWithChildren`. :) Thanks for the tip, will try it.

iAdramelk:
@C-STYR Will merge for now to unblock other PRs and then update tables separately.

PR #476, opened 2021-10-04
Add noindex in frontmatter support

iAdramelk:
This PR add two things:

1. Helper to read and parse files' frontmatter – we will need it for blog pages, case-studies pages and for the current feature.
2. Sitemap filter based on the field `noindex` inside frontmatter. 

Example usage. Add this field to the *.mdx page frontmatter section and it will not be included in the sitemap:

```
---
noindex: true
---
```


package.json
```diff
@@ -96,6 +96,7 @@
     "file-loader": "^6.2.0",
     "glob": "^7.1.7",
     "got": "11.8.2",
+    "gray-matter": "^4.0.3",
```
iAdramelk, line 99: I'm using `gray-matter` here instead if `remark` because it is a lot faster and requires a lot less boilerplate.

shashkovdanil (approved):

C-STYR (approved):

PR #491, opened 2021-10-08
Cookie Testing

C-STYR:
Ignore for now...


components/MarketoForm/helpers.ts
```diff
@@ -18,11 +19,14 @@ const getDefaultFieldValue = (
...
 
+  const cookie = parseCookies();
+
   const { value: defaultValue = "", valueFrom, parameterName } = autoFill;
 
   let value = defaultValue;
 
-  if (valueFrom === "query" && query[parameterName]) {
+  if (cookie[parameterName]) value = cookie[parameterName] as string;
+  else if (valueFrom === "query" && query[parameterName]) {
```
iAdramelk, line 29: You probably should remove `else if (valueFrom === "query" && query[parameterName])` completely if we use cookies, it will do the same thing if everything is ok and it can cause error if cookies didn't have actutal params yet for some reason.

Issue #446, opened 2021-09-21
What is Zero Trust


alexwolfe:
Current Page: https://goteleport.com/teleport/what-is-zero-trust/

Make sure the url is updated to: https://goteleport.com/what-is-zero-trust/ removing the /teleport

PR #510, opened 2021-10-12
Convert what-is-zero-trust page to Next

sandylcruz:
* Migrate 'What is Zero Trust' page to Next
* Fixes https://github.com/gravitational/next/issues/446

## Desktop
Screen Shot 2021-11-03 at 11 57 21 AM

Screen Shot 2021-11-03 at 11 57 28 AM
Screen Shot 2021-11-03 at 11 57 36 AM

## Mobile

Screen Shot 2021-10-20 at 5 33 01 PM


C-STYR (changes requested):
Some general comments: 
- Nav background and header background wave are two different images, which is different than the original
- Header background color should be a light gray gradient
- Header should have bottom border

C-STYR (approved):

PR #566, opened 2021-11-05
Update readme with instructions to add event

deliaconstantino:
- Updated step by step instructions for adding a new event through the GitHub UI.
- Removed event instructions related to terminal and code editors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,52 +158,50 @@ Get the API key via these [instructions](https://developers.google.com/youtube/v
 
 ## Marketing Team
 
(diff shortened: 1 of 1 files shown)


C-STYR (changes requested):
@deliaconstantino We don't want to remove the command line instructions, because there are some folks that will still want to refer to them; we just want to add in a separate section for the web UI. Thanks!

deliaconstantino:
@C-STYR ahh, makes sense. Added that portion back in.

C-STYR (approved):

PR #577, opened 2021-11-11
Delia/build badge component

deliaconstantino:
# Problem
Create a new `Badge` component that renders existing `Icon` component

# Solution
- Create `Badge` component that takes 2 props:
1. `name` - the named export of an SVG Icon from `components/Icon/icons.ts`
2. `size` - one of 5 possible sizes (literal union type): `"xs" | "sm" | "md" | "lg" | "xl"`

- `Badge` renders the `Icon` component, passing both of it's props into `Icon` as props.

# Example Usage
- render `Badge` component in `pages/use-cases/saas/index.mdx`:
```
<Badge iconWrapper name="check" size="xs"/>
<Section>
 <Feedback reviews={meta.reviews} />
</Section>
<Badge iconWrapper name="circleCheck" size="lg"/>
```

Output on localhost:
- checkmarks on left hand side are `Badge`
Screen Shot 2021-11-11 at 1 31 20 PM

```
<Badge name="code" size="xl"/>
<Badge name="copy" size="lg"/>
<Section>
  <Feedback reviews={meta.reviews} />
</Section>

```

Output on localhost:
Screen Shot 2021-11-11 at 1 49 55 PM

# Questions 
- do we need a color prop to render the SVG in white?

### DIFF
--- a/components/Badge/Badge.tsx
+++ b/components/Badge/Badge.tsx
@@ -1,0 +1,26 @@
+import Flex from "components/Flex";
+import Icon from "components/Icon";
+
+export type Sizes = "xs" | "sm" | "md" | "lg" | "xl";
+
+export interface BadgeProps {
+  name: "code";
+  size: Sizes;
+}
+
+export const Badge = ({ name, size }: BadgeProps) => {
+  return (
+    <Flex
+      justifyContent="center"
+      alignItems="center"
+      bg="dark-purple"
+      border="1px solid white"
+      borderRadius="circle"
+      height={48}
+      width={48}
+      boxShadow="0px 8px 16px rgba(12, 12, 14, 0.24)"
+    >
+      <Icon name={name} size={size} />
+    </Flex>
+  );
+};
--- a/components/Badge/Badge.tsx
+++ b/components/Badge/Badge.tsx
@@ -1,11 +1,12 @@
 import Flex from "components/Flex";
 import Icon from "components/Icon";
+import type { IconName } from "../Icon/types";
 
-export type Sizes = "xs" | "sm" | "md" | "lg" | "xl";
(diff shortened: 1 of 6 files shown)


components/Badge/Badge.tsx
```diff
@@ -0,0 +1,26 @@
+import Flex from "components/Flex";
+import Icon from "components/Icon";
+
+export type Sizes = "xs" | "sm" | "md" | "lg" | "xl";
+
+export interface BadgeProps {
+  name: "code";
```
sandylcruz, line 7: If name is always "code", does it need to be passed in?

components/Badge/Badge.tsx
```diff
@@ -0,0 +1,26 @@
+import Flex from "components/Flex";
+import Icon from "components/Icon";
+
+export type Sizes = "xs" | "sm" | "md" | "lg" | "xl";
```
sandylcruz, line 4: Unions typically have singular names because it can only be of one of these sizes. Maybe rename it to something more specific like `BadgeSize` in order to avoid potential name collisions.

components/Badge/Badge.tsx
sandylcruz: This can just be hard coded on line 23 if name is not changing.

components/Badge/Badge.tsx
deliaconstantino: Updated this to `BadgeSize`.

components/Badge/Badge.tsx
deliaconstantino: I looked further into this. It looks like the correct type of `IconName` can be imported into the file and used in the `Interface`. I implemented that in this commit: 3346c700ca4863b51ba59e956babdf2f7a47440f

sandylcruz (approved):

PR #580, opened 2021-11-11
update event

ardeche07:
fixes: https://github.com/gravitational/web/issues/2384


deliaconstantino:
The url link is broken. I think it's actually broken on the side of scmagazine.com. Is there an alternate url that we'd want instead?

deliaconstantino (approved):

PR #587, opened 2021-11-15
Remove mdx from components

iAdramelk:
Right now we are using mdx in some components just for metadata, which can be instead stored used existing yaml loader in a faster and more efficient way. It also make storybook setup more complicated then necessary.

So I converted these files to yaml and removed unneeded rules for webpack.


Alqanar (approved):
LGTM :)

PR #610, opened 2021-11-22
Add 54 missing storybook entries

sandylcruz:
Screen Shot 2021-11-21 at 8 22 36 PM

## Added missing entries to storybook
* Added 54 missing entries
* Modified 3 storybook entries that were present, but were not showing up on storybook due to errors
* Removed one storybook entry

## Used different storybook entry format

Screen Shot 2021-11-21 at 8 13 40 PM

* Previously `StoryComponent` was a single component, and `.bind` was being performed to make a copy of that component. Each one would have its own `.args` on it.
* `.bind` returns a `type` of `any`, which means that `.args` was not being type-checked
* In my updated format, there is a top-level `StoryComponent` and a `generateStoryComponent` function that generates a single story, which can be used any time you want to create a story. This generator returns a strictly-typed component.

Screen Shot 2021-11-21 at 8 14 07 PM

* This ensures that what is default exported from a story, matches the type `Meta` that is expected from storybook.

## Added `storybook-addon-next-router` 
* This plugin allows for components with links to render in Storybook. Previously some of these components would render an error message about being unable to map over undefined. Added use of this to `.storybook/main.js`, `.storybook/preview.js`

## Categorized components
* Resorted some components that are reusable into the root level. Moved other components into the `Site` category because they're more like reusable site components.
* Put some components in folders to improve readability. `Section` folder contains `Section`, `SectionBanner`, `SectionHeader`, `SectionPlan`, `SectionSteps`, `SectionTabs`, `SectionVideo`, and `SectionVisuals` 

## Components where entries were not added
* `Image` 
* `Head` - only injects stuff into <head> so there's no visual
* `EpisodesList` - broken, `currentItem` property is dependent on window location, therefore it doesn't work within storybook's iframe

## Components added but should be edited
* `Diagram` - image not showing up
* `BaitCard` - image not showing up


components/TextArea/TextArea.stories.tsx
```diff
@@ -1,33 +1,19 @@
...
+export default meta;
 
-Base.args = {
-  placeholder: "Placeholder",
-};
-
-export const Disabled = StoryComponent.bind({});
+export const Default = generateStoryComponent();
 
-Disabled.args = {
```
iAdramelk, line 30: I think we should not remove Disables story here.

iAdramelk (approved):
He Sandy, great work here! Thank you for adding this much stuff and also for fixing broken stories! A lot of them should have been added by me in the first place, so I'm really sorry for making you do it in my stead.

I think this PR can me merges as is, sans TextArea comments.

Some things that you may want to consider in the future. We mostly use storybook in two scenarios:

1. As a part of PRs, to demonstrate how components work without the need to launch the code locally.
2. As the documentations and display case for existing components, so we may check which components we have and reuse them instead of creating new ones.

To make these two use cases easier to implement it will be cool to have more stories for some components that shows them in different states.

E. g. for Button it would be cool to have either story, or multiple examples in the same story, that shows different sizes and shapes of the Button. This way we can check what types of buttons do we have and what types we do not. It can also help Alex and Anna with the design unification and re-usage.

Another thing what would be cool to have, but I didn't implement for a lot of components yet, are edge case scenarios. Say, some fields in the component are optional and some fields can have really long or really short content (e. g. title that takes 2 or 3 lines). It would be convenient to see how components works in these cases. This way, if something does not work, we can just show them to the design team and decide how to fix them together.

Also we does not necessary need to have storybook entry for every components, components without visual styles like Box, Flex or HeadlessButton can probably be omitted.

components/TextArea/TextArea.stories.tsx
sandylcruz: Got it, added the args back in

sandylcruz:
Thanks Alexey! I've added in TextArea's arguments and deleted the components you mentioned. I'll make a pr in the future to add stories with multiple states of components.

PR #618, opened 2021-11-22
Create ListCard and ListCardItem components

sandylcruz:
## `ListCard` and `ListCardItem`
* Add `ListCard` and `ListCardItem` components plus their storybook entries

Screen Shot 2021-11-23 at 7 56 08 AM
Screen Shot 2021-11-23 at 7 56 15 AM

## Variable description width for `SectionHeader`

Screen Shot 2021-11-23 at 8 07 18 AM
Screen Shot 2021-11-23 at 8 07 06 AM


components/SectionHeader/SectionHeader.tsx
```diff
@@ -22,6 +22,7 @@ type LinkProps = {
 export interface SectionHeaderProps {
   title: string;
   description?: React.ReactNode;
+  maxDescriptionWidth?: boolean;
```
C-STYR, line 25: Rather than a boolean prop, how about a number that defaults to 600px? That way `SectionHeader` becomes more flexible.

C-STYR (approved):
Not approved, see below.

C-STYR (changes requested):
Please don't merge, there is a problem I will explain on Slack.

C-STYR (approved):

PR #626, opened 2021-11-24
Pattern1

C-STYR:
- Extracts commonly re-used code for new design pattern
- Will eventually be used in several existing components for code simplification
- Alphabetizes messy `components/index.ts` file
- Ignore margins on screenshot within card borders, it's just example content.  This PR integrates the background, logo, links and blank white card as a wrapper
Screen Shot 2021-11-24 at 12 40 45 PM
.


components/Pattern1/constants.ts
```diff
@@ -0,0 +1,31 @@
+import webBackground from "./assets/newsletter_web.png";
+import mobBackground from "./assets/newsletter_mob.png";
+
+export type idiomClass = "purple" | "white";
```
iAdramelk, line 4: Is there a reason why types and interfaces in this file are named in `lowerCamelCase` and not in `UpperCamelCase`.

components/Pattern1/Pattern1.tsx
```diff
@@ -0,0 +1,111 @@
...
+  css({
+    display: ["none", "block"],
+    color: "rgba(255, 255, 255, 0.8)",
+    fontSize: "text-sm",
+    textAlign: "center",
+    textTransform: "uppercase",
+  })
+);
+
+const StyledLink = styled(Link)(
```
iAdramelk, line 91: Maybe it's better to create `variant` for colors inside the link component? This way we may reuse it later. Or if the colors are not used anywhere else, maybe we can replace them with one of the existing variants?

components/Pattern1/Pattern1.tsx
```diff
@@ -0,0 +1,111 @@
...
+  })
+);
+
+const StyledCopyright = styled("div")(
+  css({
+    display: ["none", "block"],
+    color: "rgba(255, 255, 255, 0.8)",
+    fontSize: "text-sm",
+    textAlign: "center",
+    textTransform: "uppercase",
```
iAdramelk, line 87: Nit: I'd move `textTransform`, `textAlign` and `fontSize` to wrapper `Box` to simplify it a little.

components/Pattern1/Pattern1.tsx
```diff
@@ -0,0 +1,111 @@
...
+            © 2021 GRAVITATIONAL, INC. ALL RIGHTS RESERVED
+          </StyledCopyright>
+          <Box
+            display={["none", "flex"]}
+            flexDirection="row"
+            justifyContent="center"
+          >
+            <StyledLink href="/tos/">TERMS OF SERVICE</StyledLink>
+            <StyledLink href="/privacy/">PRIVACY POLICY</StyledLink>
+            <StyledLink href="/security/">SECURITY POLICY</StyledLink>
```
iAdramelk, line 61: We probably should either rename these links and texts in normal case so we can change them with css later or remove `uppercase` from css. Right now it is redundant.

components/Pattern1/Pattern1.tsx
```diff
@@ -0,0 +1,111 @@
+import styled from "styled-components";
```
iAdramelk, line 1: Q: Why do we added `1` to component name? Why not just `Pattern` or `BasePattern`?

iAdramelk (approved):
Few nitpicks and questions from me.

components/Pattern1/constants.ts
C-STYR: there is not, I just wasn't aware the convention was universal.  I'll correct.

components/Pattern1/Pattern1.tsx
C-STYR: agree.

components/Pattern1/Pattern1.tsx
C-STYR: agreed

components/Pattern1/Pattern1.tsx
iAdramelk: Maybe use layout instead of pattern here then? Its more about page structure instead of design patterns. Also using `Pattern1` will require everyone to know what it is beforehand, its impossible to guess just by name. I'd name it by content instead. `LayoutForm` or something.

components/Pattern1/Pattern1.tsx
C-STYR: I'm not sure where this color originated, tbh.  I think it came from /web.  I have replaced it with one from `theme.ts` and created a variant.

components/Pattern1/Pattern1.tsx
C-STYR: Layouts and page structures are design patterns; also, we have lots of elements named some variation of "layout" but none named "pattern".  Semantic names only go so far in any case. How does one distinguish between `SectionVisuals` and `SectionPlan` without knowing what they look like beforehand? It's impossible to guess, which is why we have Storybook.  If it becomes a problem, it can always be renamed.

PR #634, opened 2021-11-26
Blog article layout

Alqanar:
Article layout
layout

Search
https://user-images.githubusercontent.com/41822696/146043776-025b36ea-9b13-4724-b7da-4073c621338d.mov


pages/api/search/index.ts
```diff
@@ -0,0 +1,54 @@
...
+const RESOURCES_URL = /(\/resources\/)/;
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  let rawResults = [];
+
+  try {
+    if (BLOG_URL.test(req.query.url as string)) {
```
iAdramelk, line 20: Q: Why send `url` as a param? Is it not a part of `req`?

pages/api/search/index.ts
```diff
@@ -0,0 +1,54 @@
+import type { NextApiRequest, NextApiResponse } from "next";
```
iAdramelk, line 1: Nit: Can you please add some comments to this file, describing why we do different stuff depending on url, etc. I know the context of the task, but then somebody will need to change something here after we finish the project, it will be hard for them to understand what's going on.

server/search-api-helpers.ts
```diff
@@ -0,0 +1,57 @@
+const SITE_INDEX = process.env.ALGOLIA_SITE_INDEX_NAME;
+const BLOG_INDEX = process.env.ALGOLIA_BLOG_INDEX_NAME;
+const RESOURCES_INDEX = process.env.ALGOLIA_RESOURCES_INDEX_NAME;
+
+export type indexName = "blog" | "site" | "resources";
+
```
iAdramelk, line 6: Nit: Can you please add comment with the link of the docs of corresponding algolia docs for query params.

layouts/BlogArticle/BlogArticle.tsx
```diff
@@ -0,0 +1,194 @@
...
+import styled from "styled-components";
+import css from "@styled-system/css";
+import Drift from "react-driftjs";
+import { format, compareDesc } from "date-fns";
+import NextImage, { ImageProps } from "next/image";
+import { components as baseComponents } from "layouts/SitePage";
+import Box from "components/Box";
+import Flex from "components/Flex";
+import FeaturedArticleCards from "components/FeaturedArticleCards";
+import type { BlogArticle as BlogArticleInfo } from "components/FeaturedArticleCards/types";
```
iAdramelk, line 11: Nit: We may want to move `FeaturedArticleCards` and `BlogArticle` inside `layout/BlogArticle` folder if we don't plan to use them anywhere else. We also may skip creating separate storybook pages for them this way. :)

server/resources-helpers.ts
```diff
@@ -58,3 +59,24 @@ export const getResourcesData = (type: string) => {
 
   return resourcesPageInfo;
 };
+
+export const getArticlesList = () => {
```
iAdramelk, line 63: Let's add `limit` param to the number or returned results here. Original blog had more than 180 articles and we only need to show last three. We can just return 3 here, instead of loading all 180 for each page.

layouts/BlogArticle/BlogArticle.tsx
```diff
@@ -0,0 +1,194 @@
...
+    .sort((a, b) =>
+      compareDesc(new Date(a.frontmatter.date), new Date(b.frontmatter.date))
+    )
+    .slice(0, 3);
+
+  const rawAllTags: Set<string> = new Set();
+  articlesList?.forEach((article) =>
+    article.frontmatter.tags.forEach((tag) => rawAllTags.add(tag))
+  );
+  const allTags = Array.from(rawAllTags);
```
iAdramelk, line 103: Nit: Let's move tag filtering to the server. Original blog had more that 180 articles, we probably don't want to load them all to the client just to get tag list.

iAdramelk (approved):
I added a bunch of nitpicks. you can merge the brunch without addressing them because they don't stop us from starting posts migration, but please address them in separate PR.

components/Tags/Tags.tsx
```diff
@@ -0,0 +1,87 @@
+import { useCallback, MouseEvent } from "react";
```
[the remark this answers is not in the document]
Alqanar: Ok, I'll add this component to the storybook later

components/Tags/Tags.tsx
```diff
@@ -0,0 +1,87 @@
...
+  tags: string[];
+  size?: Size;
+} & BoxProps;
+
+export function Tags({ tags, size, ...props }: TagsProps) {
+  const handleClick = useCallback((e: MouseEvent<HTMLButtonElement>) => {
+    e.preventDefault();
+
+    const button = e.target as HTMLButtonElement;
+    console.log(button.innerText);
```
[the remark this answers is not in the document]
Alqanar: No, I didn't add a comment, that will be added later

pages/api/search/index.ts
Alqanar: A: I send the url as a parameter to know from which page the request was made.

components/EmailSubscribe/EmailSubscribe.tsx
```diff
@@ -1,10 +1,24 @@
+import { Fragment } from "react";
```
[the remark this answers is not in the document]
Alqanar: I think it's better to keep them in one place. The components are not very big and it will be easier to find them when you need them.

PR #640, opened 2021-12-01
Fix storybook entry

sandylcruz:
* Fix `TOCLayout` storybook entry where text was overlapping

Screen Shot 2021-11-30 at 4 05 33 PM


C-STYR:
This component does not need a storybook entry, we can git rid of it.

PR #642, opened 2021-12-01
Fix components' imports

iAdramelk:
Imports using `{} from "components";` syntax were preventing me from reusing `OurOffices` component in the new layout (TypeScript breaks on recursive imports). So I fixed it and also fixes other similar cases in other files to prevent future problems.

This review request just FYI. I will merge it myself if build does not return errors.


iAdramelk:
> So the key point here is to import components individually, correct?

Its more that we can't have circular dependencies: We can't import `Box` from `components/index.ts`, but then also add current component to `components/index.ts` because it will create circular dependency which will break TypeScript.

PR #645, opened 2021-12-01
Delia/add enterprise signup

deliaconstantino:
# Add `/signup/enterprise` page to Next
- matches existing https://goteleport.com/signup/enterprise page except uses a white background
- preview [here](https://teleport-docs-next-git-delia-add-enterpris-6c04c0-gravitational.vercel.app/signup/enterprise/)

## Questions:
- incorporated existing `EnterpriseSignup` component with Marketo form. 
  - __Note:__ Overrode some styling on `EnterpriseSignup` component to create a matching page to the `web` version. However, `EnterpriseSignup` is not currently used anywhere else in the `next` repo. Do we want to make those changes to the `EnterpriseSignup` component itself?

## Demo:
https://user-images.githubusercontent.com/70108137/144308731-51ad6ac0-12eb-465c-9def-d0a4698ded7e.mov


pages/signup/enterprise.tsx
```diff
@@ -0,0 +1,98 @@
...
+          mt="auto"
+        />
+      </RightSide>
+    </Pattern1>
+  );
+};
+
+const LeftSide = styled(Flex)(
+  css({
+    display: "flex",
```
sandylcruz, line 77: `Flex` is already `display: "flex"` so you don't need this

pages/signup/enterprise.tsx
```diff
@@ -0,0 +1,98 @@
...
+    backgroundColor: "#F6F7F8",
+    flexBasis: ["100%", "35%"],
+    padding: [2, 4],
+  })
+);
+
+const RightSide = styled(Flex)(
+  css({
+    flexDirection: "column",
+    display: "flex",
```
sandylcruz, line 90: don't need this `display: flex` either

sandylcruz (approved):

pages/signup/enterprise.tsx
deliaconstantino: ah good point, thank you!

PR #648, opened 2021-12-02
Press page migration

iAdramelk:
Port of https://goteleport.com/about/press/

Some caveats:

- I tried to make design closer to our existing one. Because of that header section have different padding, list of news have gradient background instead of pattern (our existing patters look terrible here) and press blocks at the bottom have unified styles and smaller line height for text.
- A lot of news articles on the original site have `publication: Security Weekly`, even if they are not published there. I blame copy-paste here, but decided not to fix them for now and ask if they are needed to be fixed first.
- Original images were somewhat a mess with duplicate logos, different locations, and stange naming conventions. I fixed issues with naming, locations and duplicates, but we still have some quality problems with some logos. E. g. `security-weekly.png` is only 121x69 in size. We probably should replace them with better quality versions.

Demo: https://teleport-docs-next-nptyxi5kh-gravitational.vercel.app/about/press/

CC: @alexwolfe Please check the list above, I need you input about what to do with this stuff.


layouts/PressPage/PressList.tsx
```diff
@@ -0,0 +1,39 @@
+import styled from "styled-components";
```
iAdramelk, line 1: I put `PressList` and `PressListItem` near the layout instead of `components` folder because we did not reuse them. This is also the reason I does not create storybook page for them.

pages/about/press/index.mdx
```diff
@@ -0,0 +1,343 @@
+---
+layout: press
+title: Teleport in the news
+h1: Recent publications and media resources
+description: The list of recent publications and media resources
+articles:
+  - title: Overcoming The Complications Of Access In Cloud With Teleport
+    description: Why did Ev Kontsevoy, Co-Founder and CEO at Teleport, create Access Plane?
+    publication: Security Weekly
```
iAdramelk, line 9: @alexwolfe Take note that this (and some others below) `publication` value is wrong, but it was this this way on the old site. Should I fix them?

C-STYR (approved):
I'm approving the code only here.  @alexwolfe will need to OK the stylistic variations you mentioned above...

layouts/PressPage/PressPage.tsx
```diff
@@ -0,0 +1,158 @@
...
+        <OurOffices />
+        <TryTeleport />
+      </Layout>
+      <Footer />
+      <Drift appId={process.env.NEXT_PUBLIC_DRIFT_ID} />
+    </>
+  );
+};
+
+const StyledHeader = styled("h3")(
```
Alqanar, line 110:
```suggestion
const StyledHeader = styled("h2")(
```

Maybe using the `h2` tag would be more correct here?

Alqanar (approved):
LGTM!
Take a look a few comments, pls

alexwolfe:
Design wise it looks good, I did notice a issue on mobile that needs to be fixed though: 

image

PR #658, opened 2021-12-06
Removed "our vision"

halyna-y:
We no longer need this section so I removed it.


C-STYR:
Bit of a problem here: 
Screen Shot 2021-12-07 at 10 25 03 AM

Once the `MessageBanner` component was removed, two sections with purple backgrounds became adjacent.  May I suggest moving the `Section` containing `Feedback` between `WorksWith` and `EasyStart` like so? 
Screen Shot 2021-12-07 at 10 28 19 AM

halyna-y:
@C-STYR Updated, please take a look.

C-STYR (approved):
Well done Halyna!

PR #668, opened 2021-12-08
Cleanup utils/url

iAdramelk:
This PR have multiple goals:

1. To remove all imports from the `utils/url.ts`. We need to use it both on server on client, but we also want to convert server to esm. Esm can't work with `.json` so we need to load it with `fileReadSync` now,  which does not work on client, so I have to remove loading `config.json` from it.
2. To simplify docs code. Current code have some strange custom logic and checks that required data from both client and server to work and logic was scattered to different places. I simplified it a little and tried to move all logic together.
3. To rename method names and add basic docs to utils while I'm at it.

Main changes:

1. More data in the docs pages context. Now it have info about `current`, `latest` and all `available` versions.
2. New hook `useDocPath` that encapsulates logic of nirmalizing docs url.
3. `<Head>` now have an optional `url` param that will be used instead of `react.asPath` for building canonical urls.

Update:

- Reworked hooks. Now we have two global ones:
  - `useCurrentHref` - will return current href with resolved rewrites.
  - `useNormalizedHref(href)` - will return site's absolute href with correct rewrites and scope for docs.
- Removed `DocsLink` and bunch of logic from `Link` because we does not need them anymore.


C-STYR (approved):

Alqanar (approved):
LGTM!

PR #669, opened 2021-12-08
Implement redesigned guides

sandylcruz:
# Implement redesigned version of guides

* Migrate individual pages e.g. https://goteleport.com/resources/guides/teleport-digitalocean-1-click-droplet/
* Preview e.g. https://teleport-docs-next-git-sandy-guides-gravitational.vercel.app/resources/guides/teleport-digitalocean-1-click-droplet/

## Overall changes made

* Because there is not yet a design for the mobile version, I kept similar overall styling/hover state as the previous version
* Modified `Video` to take `borderRadius` prop  instead of `hasRoundedEdges` with options of `sm`, `md`, `lg`, `none`, with default of `md`

## `VideoRow` changes

* Changed `VideoColumn` to `VideoRow`, updated styling
* ~~On mobile, `VideoRow` is `display: none`. I tried changing its `order` property so that `VideoRow` would display underneath the text (as it was in the previous version). However because the text and `VideoRow` have different parents, this doesn't work.~~
* Updated mobile design where `VideoRow` displays initially as a row of 4 videos on desktop screens, then displays as 2 columns of 2 videos, and finally displays as one column on mobile screens

mobile-behavior

* `VideoRow` displays the following:
    * If there are no associated labels to a guide, `VideoRow` will display the 4 most recent videos
    * If there are guide labels and there are 4+ videos with matching labels, will display the most recent videos with matching labels only 
    * If there are guide labels and there are NOT 4 videos with matching labels, will display the videos with matching labels plus the most recent videos.`VideoRow` will not display a duplicate video if one of the most recent videos also shares a matching label.

Screen Shot 2021-12-10 at 3 42 20 PM
Screen Shot 2021-12-10 at 3 43 51 PM


C-STYR:
@sandylcruz a couple things need to happen before we can proceed with this redesign:

1. Please figure out why the OG images for the videos in the `VideoColumn/VideoRow` component have degraded quality. I don't believe the original server from `/web` had this problem but unable to verify now
2. We need @alexwolfe approval to hide `VideoRow` on mobile.  I personally think it should remain and the display logic shouldn't be too challenging.

Otherwise, nice job.  Your skills are definitely sharpening!

components/VideoRow/VideoRow.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+export const VideoItem = ({ author, href, title, videoId }: VideoItem) => {
+  return (
+    <Link
+      href={href}
+      mr={5}
+      textDecoration="none"
+      css={css({
+        "&:last-child": {
+          mr: ["", 0],
```
C-STYR, line 94: instead of passing an empty string in a responsive styling array, pass null and Styled System will not produce unnecessary CSS, e.g. `mr: [null, 0]`

C-STYR (approved):
Looks good with a couple comments, and also the margin between video link cells and main page video needs to be fixed below the smallest breakpoint:
Screen Shot 2021-12-16 at 10 38 49 AM

PR #672, opened 2021-12-09
Added two speakers bio

halyna-y:
I added two more speakers for March and April. Will need to update once we have titles and abstracts from them.

(From Cole):
Preview [HERE](https://teleport-docs-next-git-halyna-addmichaelkeren-gravitational.vercel.app/security-visionaries-2022/)


C-STYR:
@halyna-y Hey, really nice job. Just a couple issues, then this can be approved: 
The first is that I only see one file change:
Screen Shot 2021-12-09 at 9 17 15 AM
But it looks like you have attempted to add 2 images because you have referenced them in the code. So in total there should be 3 files changed (2 new images plus the .mdx file itself).  Not sure if you're using the Web UI or the command line, but it looks like the images didn't make it into the commit. 
Also, you need to make sure to check the preview (unless you're running the file locally, in which case see below).  If so, you would have seen this: 
Screen Shot 2021-12-09 at 9 16 20 AM
The auto-generated preview, which appears like this: [https://teleport-docs-next-git-halyna-addmichaelkeren-gravitational.vercel.app](https://teleport-docs-next-git-halyna-addmichaelkeren-gravitational.vercel.app), needs a slight modification in order to work.  You need to add the end of the url path to it.  In this case, it is `/security-visionaries-2022/`. So now if you put [https://teleport-docs-next-git-halyna-addmichaelkeren-gravitational.vercel.app/security-visionaries-2022/](https://teleport-docs-next-git-halyna-addmichaelkeren-gravitational.vercel.app/security-visionaries-2022/) into a new tab, you'll see what you need to see.  It's also customary to take this whole preview url and put it at the top of the page in the first comment so others reviewing don't have to add it.  The way to do that is with markdown syntax:
`[the visible link goes here](the url goes here)`
The dev team would do something like this: 
`Preview [HERE](https://teleport-docs-next-git-halyna-addmichaelkeren-gravitational.vercel.app/security-visionaries-2022/)`.  I've done it for you above. If you edit the comment you should be able to see the raw text. 
OK, if you were running the file locally using `localhost:3000/`, then you just need to add the final bit of the path to that url: `localhost:3000/security-visionaries-2022/`.  But you'd still need to add the preview link in the top comments. 
Right, that's a lot to think about so I'll leave it there. If you need help fixing this today I'm happy to do so.

C-STYR:
@halyna-y I added Anya's edits to this PR and removed the bad links for the images, replacing them with those of Troy (you'll need to replace those).

halyna-y:
@C-STYR I don't know how to add images, I made a separate request for that but happy to learn if you can show me. https://github.com/gravitational/web/issues/2513

alexwolfe (approved):

PR #679, opened 2021-12-14
Make latest docs version 8

benarent:
This PR links v8 docs to `branch/v8` and creates v9 which links to master. 

This PR should be merged alongside the PR release around 6am pacific.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,4 +25,8 @@
 [submodule "content/8.0"]
 	path = content/8.0
 	url = https://github.com/gravitational/teleport.git
+	branch = branch/v8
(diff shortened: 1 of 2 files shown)


wadells (approved):
LGTM.

You may want to update the hash at `content/v8` as swapping the submodule from `master` to `branch/v8` it isn't a fast forward.  I don't know that this would break anything, but better to be safe than sorry.

benarent:
> You may want to update the hash at `content/v8` as swapping the submodule from `master` to `branch/v8` it isn't a fast forward

I tried following the steps. I'll loop back to this after launch.

PR #684, opened 2021-12-14
Feature/blog rss

iAdramelk:
Ok, this PR ended up larger that I expected, but I split it to separate commits to make it easier to review.

*1. [Add universal sorter function](https://github.com/gravitational/next/pull/684/commits/280228ba7a78ca094669688b67afcb3cf34bb7b6) commit*

We sort by dates and sort by `ASC`/`DESC` a lot, so I created a function that automates comparison.

Use it like that:

```
sort(valueA, valueB, "DESC");
```

See tests for examples.

*2. [Add sort, filter and better types to pages-helpers](https://github.com/gravitational/next/pull/684/commits/c7e2bf8d1cfc2489fefd69fa552df72b1fd9e10b) commit*

We also do a lot of filtering and sorting of the `getPagesInfo` results. So I added them inside the function itself with a better types support.

Example:

```
getPagesInfo<BlogPostFrontmatter>(`blog/**/*.mdx`, {
    filter: ({ title }) => Boolean(title),
    sort: "date",
    order: "DESC",
    limit: 20
  });
```

This will filter all blog posts with title field in frontmatter sort them by "date" and "DESC" and return first 20 of them.

*3. [Simple RSS generator](https://github.com/gravitational/next/pull/684/commits/95ccdb0ae3118602cfb9eb4a532f3b3d1aeda232) commit*

This will take frontmatter from the posts, create simple rss feed from it (no actual post content, just description from metadata) and save it to `/public/blog/index.xml`.

Sasha said that we should skip generating actual html for content for now because it is too time consuming.


C-STYR (approved):
check linter issue, otherwise OK

PR #704, opened 2021-12-22
Refactor AnnouncementBar

iAdramelk:
`AnnouncementBar` is using `svgr` components for logos right now, but we don't really need them as react components because we do not modify them with css in any way. So I replaces them with static images.

It reduced a size of the pages that used this component a lot. 

E. g. `/what-is-zero-trust/` before refactor – `872kb`, after – `255kb`.

Update:

Ok. Its hard to stop after you start refactoring stuff. Form the first review:

- Cleaned up and restructured all svg logos (some manual file editing make it easier for svgo to compress).
- Replaced all but 2 of png logos with svg versions. New `darktrace` and `draft-kings` are flat without gradient, but they are used on their official sites without gradients so it should be ok.
- Optimized size of the 2 pngs that left (they are not 2/3 smaller).
- Added separate color version of `samsung` logo (bw version was used before which is not totally correct, because samsung brand book says to use blue color).
- Refactored components and reduced repeated code.
- Changed component API to the more compact and universal one.
- Component can now work with any number of rows.
- Animation is now paused then you hover the logos row.
- Added storybook page with all combinations of component that we actually use to make testing easier.
- Fixed size and placement of logos in the desktop version.
- Reworked sizes and spacings of elements on mobile because they looked not very good and consistent.


pages/index.mdx
```diff
@@ -210,7 +211,6 @@ $ helm install
 <Section bg="flatWhite" px={10} py={10}>
   <AnnouncementBar
     caseStudy
-    title
```
iAdramelk, line 213: `title` prop was not actually used anywhere in the component.

pages/use-cases/aws/index.mdx
```diff
@@ -41,18 +41,12 @@ import largorgs from "./assets/for-largorgs.svg";
...
-    <Link
-      href="https://aws.amazon.com/marketplace/pp/prodview-syjxiiw25cr3u?sr=0-1&ref_=beagle&applicationId=AWSMPContessa"
-    >
+    Teleport is available in the
+    <Link href="https://aws.amazon.com/marketplace/pp/prodview-syjxiiw25cr3u?sr=0-1&ref_=beagle&applicationId=AWSMPContessa">
       <Image ml="10px" src="./assets/aws-marketplace-logo.svg" />
     </Link>
-</NewsItem>
-  <AnnouncementBar
-    backgroundColor="lightest-gray"
```
iAdramelk, line 52: `backgroundColor` and other css props does not affect components in any way right now.

Alqanar (approved):

C-STYR (changes requested):
Just so this doesn't accidentally get merged: this refactor is incomplete until presentation is brought in line with the original:
- grayscale ABar is the wrong color and the spacing between logos is incorrect
- color version is not displaying all logos on several pages
- color version logos are in some cases the wrong size

Alqanar (approved):

C-STYR:
@iAdramelk I'm sure you've got more important things to work on in your remaining time.  This component still needs additional refactoring - the animation timing function is off and the spacings still need work (see top instance on the homepage in the preview).  I'm going to assign Delia or Sandy to clean this up in January so they can have some experience with animations.  Thanks for your hard work on this, we'll take it from here.

C-STYR (approved):

PR #715, opened 2021-12-28
Remove components/index.ts

iAdramelk:
All components used `components/index.ts` ware combinedin the common chunk by next.js, which were increasing first page load for every page by ~60-70kb of javascript. I replaced it with separate import to reduce page sizes.


C-STYR:
Looks good, but let's hold off merging until 4 January when most of us are back at work.  I want to make sure we can respond to breakages quickly.  Thanks Alexey.

C-STYR (approved):

PR #773, opened 2022-01-14
`<WhyTeleport/>` Component for homepage

deliaconstantino:
# Create `<WhyTeleport/>` Component for homepage

## Implement animation initiation based on scroll:
Use [`useWindowScroll`](https://github.com/streamich/react-use/blob/master/docs/useWindowScroll.md) hook from [`react-use`](https://github.com/streamich/react-use) dependency, which is already added to this repo. In this use-case, this monitors the y-axis for when to initiate the animation component. The animation is initiated when the component scrolls into view from the top.

## Animation sizing:
Implemented animation at smaller sizes for mobile and midway (between 900px-1200px) breakpoints. Because each image that creates the animation is positioned absolutely, specific sizes and locations must be provided for each image at all sizes.
- mobile size: 50% of Figma model
- 900-1200px size: 80% of Figma model

## Notes
1.  may want to update timing on animation. This is a very simple change; can update the animation delay as needed.

2. this is the MVP animation with:
  -  arm raise
  - button press
  - screens sliding into place
  
 Can also add additional elements we discussed such as spinning button, etc

3. Used [NextImage](https://nextjs.org/docs/api-reference/next/image) for all images. A warning renders in the console about using `layout="fill"` and absolute positioning of images. I used `layout="fill"` for images in order to place each image inside containers with specific sizes and positions that respond to screen size shifts. I was unable to find a solution using NextImage without the warning. However they may a better solution using `srcSet`, per [this section](https://nextjs.org/docs/api-reference/next/image#layout) of the docs

4. This component could be be assembled as its parts on the new homepage, rather than imported as `<WhyTeleport />`. Created component file as a placeholder for now.

### Preview [here](https://teleport-docs-next-git-delia-homepage-secu-84fadf-gravitational.vercel.app/) 
scroll to "Security Software Done Right?" section

## Blockers:
- added `<WhyTeleport />` component to homepage for testing/demonstration. Remove from homepage before merging.


components/SectionVisuals/SectionVisuals.tsx
```diff
@@ -7,56 +7,77 @@ import { P } from "components/MDX";
 
 export interface SectionVisualsProps {
   subtitle?: string;
-  title: string;
+  title?: string;
```
C-STYR, line 10: When would this component not have a title? It has one in every instance in the codebase.

components/SectionVisuals/SectionVisuals.tsx
```diff
@@ -7,56 +7,77 @@ import { P } from "components/MDX";
...
   contentWidth?: string;
   children: React.ReactNode;
   description: React.ReactNode;
+  imageLeft?: boolean;
+  smallerTitleFont?: boolean;
```
C-STYR, line 15: Creating boolean props like this is bad practice in scenarios where there are more than two options. Instead you should create a way for styles to be passed in directly.

components/SectionVisuals/SectionVisuals.tsx
```diff
@@ -7,56 +7,77 @@ import { P } from "components/MDX";
...
+            })}
+          >
+            {description}
+          </Box>
+        )}
       </Box>
       <Flex
         flex="0 0 auto"
         alignItems="center"
+        justifyContent={imageCenteredOnMobile && "center"}
```
C-STYR, line 80: If `imageCenteredOnMobile` is false, what does this expression evaluate as? Is that a legitimate value for `justifyContent`? How should a responsive style for this (and other) prop(s) look?

components/SectionVisuals/SectionVisuals.tsx
```diff
@@ -7,56 +7,77 @@ import { P } from "components/MDX";
...
   subtitle?: string;
-  title: string;
+  title?: string;
   contentWidth?: string;
   children: React.ReactNode;
   description: React.ReactNode;
+  imageLeft?: boolean;
+  smallerTitleFont?: boolean;
+  imageCenteredOnMobile?: boolean;
+  provideDescriptionStyling?: boolean;
```
C-STYR, line 17: Same as above. If there are too many hard-coded styles, you need to fix each case individually with a more generally-styled component rather than add further complication.  The styles within the `<Box/>` on lines 56 to 74 should be moved to their specific instances and `provideDescriptionStyling` should go away.

components/WhyTeleport/WhyTeleport.tsx
```diff
@@ -0,0 +1,66 @@
...
+        description={
+          <Box
+            flexShrink={2}
+            boxSizing="border-box"
+            width={["95vw", "100%", "480px"]}
+            mx="auto"
+            mt={[4, 4, 7]}
+          >
+            <StyledText>{text}</StyledText>
+            <Button
```
C-STYR, line 38: What do other buttons in this codebase look like on mobile?

components/SectionVisuals/SectionVisuals.tsx
```diff
@@ -7,56 +7,77 @@ import { P } from "components/MDX";
...
-            li: {
-              color: "black",
-              fontSize: "text-md",
-              lineHeight: "md",
-            },
-          })}
-        >
-          {description}
-        </Box>
+        {title && (
```
C-STYR, line 42: See above comment re `title`

components/WhyTeleport/PamAnimation.tsx
```diff
@@ -0,0 +1,284 @@
...
+    left: 477,
+    height: 63,
+    width: 27,
+    zIndex: "3",
+    src: commentLine,
+    alt: "comment line screen",
+  },
+];
+
+export interface AnimatedImageContainerProps {
```
C-STYR, line 84: why is this exported?

components/WhyTeleport/PamAnimation.tsx
```diff
@@ -0,0 +1,284 @@
...
+  top: number;
+  left: number;
+  height: number;
+  width: number;
+  src: string;
+  alt: string;
+  zIndex: string;
+}
+
+export const AnimatedImageContainer = ({
```
C-STYR, line 95: why is this exported?

C-STYR (changes requested):
Overall, very impressed with your animation; I thought that was the bit you would struggle with but you've done well.  However it's actually the modification of `SectionVisuals` that needs significant revision.  Either use my comments here to de-clutter `SectionVisuals` or create a custom component.

components/SectionVisuals/SectionVisuals.tsx
```diff
@@ -7,56 +7,77 @@ import { P } from "components/MDX";
...
+  imageCenteredOnMobile = false,
+  provideDescriptionStyling = false,
 }: SectionVisualsProps) => {
   return (
     <Centrator
       alignItems="stretch"
-      flexDirection={["column", "row"]}
+      flexDirection={
+        imageLeft === true
+          ? ["column-reverse", "row-reverse"]
```
C-STYR, line 36: The animation should probably go on the bottom in this component - even if the image is on the left.  But once again, these boolean props should probably be refactored.

components/WhyTeleport/WhyTeleport.tsx
deliaconstantino: Good point, updated to `width={["100%", "auto"]}` to match codebase

components/SectionVisuals/SectionVisuals.tsx
deliaconstantino: I believe Alex mentioned that the text should slide under the animation for the mobile version. I will double-check on that.

components/SectionVisuals/SectionVisuals.tsx
deliaconstantino: Good point, updated title

components/SectionVisuals/SectionVisuals.tsx
deliaconstantino: updated these to be passed in directly

components/SectionVisuals/SectionVisuals.tsx
deliaconstantino: updated to `imageLocation` prop, implementing these suggestions, with this [commit](https://github.com/gravitational/next/pull/773/commits/4e6a4da18510e9efad514d0b70957711fad9b939)

components/SectionVisuals/SectionVisuals.tsx
deliaconstantino: add a `secondary` variant for these styles and removed `provideDescriptionStyling` prop

components/SectionVisuals/SectionVisuals.tsx
deliaconstantino: fixed

components/WhyTeleport/PamAnimation.tsx
deliaconstantino: removed this export

components/WhyTeleport/PamAnimation.tsx
deliaconstantino: removed this export, and other unused exports

PR #776, opened 2022-01-18
Fix recaptcha erro in esm build

iAdramelk:
P. S. Only check the second commit. First one reverts old commits, third one fixes build error in `main`.


utils/recaptcha.ts
```diff
@@ -26,6 +26,32 @@ export const load = (): Promise<void> => {
   });
 };
 
+interface CleanupOptions {
+  ID: number;
+  containerID: string;
+  callbackFNName: string;
```
C-STYR, line 32: why not just `callback`?

utils/recaptcha.ts
```diff
@@ -26,6 +26,32 @@ export const load = (): Promise<void> => {
...
+}
+
+// Recaptcha unmount function
+
+const cleanup = ({ ID, containerID, callbackFNName }: CleanupOptions) => {
+  const el = document.getElementById(containerID);
+
+  // reCaptcha will leave some DOM elements inside tagrt node even after reset.
```
C-STYR, line 40: Can you please provide further comments about the overall purpose of this code.  Why is it important that reCaptcha does or does not leave DOM elements after reset?

utils/recaptcha.ts
```diff
@@ -52,9 +78,23 @@ export const useRecaptcha = (UID: string) => {
       return;
     }
 
+    // Because load and init are async we can end in the state then
+    // we unmounted component before they are finished,
+    // "isMounted" if a flag that helps us check it
```
C-STYR, line 83:
```suggestion
// Because load and init are async we can end in the state where
// a component is unmounted before it is finished;
// "isMounted" is a flag that helps us check it
```

C-STYR (approved):

utils/recaptcha.ts
iAdramelk: I just reuse name form the main hook here. And it is named this way here because it not callback itself, but a string with the name of the key in the `window` object we will assign real callback later.

We can't assign CB directly to recaptcha because it does not work like than. We need to give it a callback function name that it can call later.

utils/recaptcha.ts
iAdramelk: It's all related to the `useEffect` hook and component's lifecycle:

1. React component is mounted and dom for it is rendered.
2. After that `useEffect` code is called.
3. Inside `useEffect` we use some async code that is not blocking execution. E. g. we asynchronously init recaptcha code and then set recaptcha's id to the component's state.
4. But component can be not only mounted, but also unmounted is some situations. And then mounted again.

So now we have two potential problems:

1. We mounted component, started recaptcha init process and then unmounted it before it was finished. In this case recaptcha'a callback will throw an error because it can't set ID to state because state does not exist anymore.
2. If we mount the component again after we unmounted it before and recaptcha's instance is already exists, our code will break.

So we need to do the following:

1. Remove recaptcha instance and everything related to in unmount. Unforunately `recaptcha.reset()` does not remove everything and we need to remove dom elemenets manually.
2. If unmount happens before the recaptcha instance was fully initialized, prevent remounting and cleanup junk.

Alqanar (approved):

utils/recaptcha.ts
C-STYR: @iAdramelk thanks for this.  I realize I wasn't being very clear, sorry, so I'll refactor:  by "comments" I meant comments in the codebase, so that everyone can refer to them long after this PR is merged. Your first section above is probably not needed, but a concise version of the second and third sections would be helpful.

iAdramelk: Sure, updated the comments.

PR #779, opened 2022-01-18
Fix parsing of event dates and retrieval of events

sandylcruz:
* Fixes https://github.com/gravitational/web/issues/2653
* Preview [here](https://teleport-docs-next-git-sandy-webinar-gravitational.vercel.app/about/events/)

Screen Shot 2022-01-19 at 8 51 34 AM

# Fix formatting of event dates

- Previously event dates were off by one day due to dates being parsed in UTC time zone (8 hours ahead) automatically. To fix this, I wrote the event times in their equivalent  UTC format:
     - Before: `2022-03-13`
     - After: `2022-03-13T08:01:00Z`
- The time is set to `08:00:01` because that's the equivalent UTC time to 12:01 am PST.

# Fix retrieval of events

- The events page was only returning events between two days from now up to 30 days from now (e.g. between 01/20/2022 - 02/17/2022 ) due to an offset of 2 days that was written into the code
- Additionally, once the above was fixed, events for the same day were not being retrieved. I fixed this by changing what `now` points to. 
- Before, it was pointing to the exact time right now(i.e. `Wed Jan 19 2022 09:04:54 GMT-0800 (Pacific Standard Time)`)
- I changed it to point to today's date at midnight (i.e. `Wed Jan 19 2022 00:00:00 GMT-0800 (Pacific Standard Time)`) in order to facilitate retrieval of today's events.

## Misc
* Removed old event entries
* Added missing new webinars to the webinars list on the events page


C-STYR (commented):
Good, but we need to retain the YYYY-MM-DD format so that the marketing folks can have as few barriers as possible to adding events.  Can you please create a helper function that adds the `T08:01:00Z` portion before at or before sort?

sandylcruz:
@C-STYR When dates are parsed from MDX, they automatically are formatted to the equivalent UTC date/time if you're not explicitly stating a time so adding a helper functions like that didn't work. Figured out another way to get the correct dates shown though.

C-STYR (approved):

PR #819, opened 2022-01-25
Fix new lines in code blocks

### LINKED ISSUE #811: Escaped newlines in ```code snippets rendering as spaces
[For example](https://goteleport.com/docs/kubernetes-access/helm/guides/aws/#step-5-set-values-to-configure-the-cluster):

image

Since we're moving code snippets toward the `code` format, this means that a lot of commands with flags that span multiple lines (e.g., `helm install` commands) render unexpectedly.

iAdramelk:
Fix #811

At some point of time we lost `display: block` on `span`s in the custom code blocks. It's probably was set by some if it's parent.

I fixed it for now by using `> span` selector, but to make it more bulletproof its better to convert lines in the components itself in `remark-plugin`. CC @Alqanar 

Update:

- Replaced tags inside the plugin with React components.
- Removed non-exisitng mdx nodes from types (they were removed in rc2).

Update 2:

- @Alqanar can you please check updated test fixtures, I'm not totally sure we need nested spans in command lines.


C-STYR:
@iAdramelk let's go with your "bulletproof" solution rather than merging this one then.  Can you take care of it please?

C-STYR (approved):

iAdramelk:
@Alqanar con you please check the last comment, I removed spans inside `p` tags and moved `[data-type='descr']` to work on every comment node.

PR #896, opened 2022-02-07
Restore Form Submission Tracking in `/signup`

C-STYR:
Partially addresses gravitational/web#2719.

The original script [found here](https://github.com/gravitational/web/blob/master/layouts/signup/list.html) contained a `dataLayer` push in an optional `onSuccess` callback: 
```
<script>
    window.teleport.renderSignupRequestForm(document.getElementById('app'), {
      apiUrl: 'https://teleport.sh', //https://cloud.gravitational.io
      onSuccess: () => {
        window.dataLayer = window.dataLayer || [];
        window.dataLayer.push({
          'event': 'formSubmit',
          'formType': 'Teleport Pro Signup',
        });
      }
    });
  </script>
```
This PR adds that push to the `/next` version.


utils/cloud-signup.ts
```diff
@@ -25,14 +25,20 @@ function loadLib(): Promise<typeof window.teleport> {
...
-export async function renderSignupRequestForm(element: Element): Promise<void> {
+interface optionsInt {
+  apiUrl: string;
+  onSuccess: () => void;
+}
+
+export async function renderSignupRequestForm(
+  element: Element,
+  options: optionsInt
+): Promise<void> {
```
iAdramelk, line 36: Do we plan to reuse this form somewhere else? If no, you an just hardcode onSuccess inside without using options.

iAdramelk (changes requested):

components/CloudSignup/CloudSignup.tsx
```diff
@@ -8,8 +8,21 @@ export default function CloudSignup(props: BoxProps) {
...
+    const optionsObj = {
+      apiUrl: API_URL,
+      onSuccess: () => {
+        window.dataLayer = window.dataLayer || [];
+        window.dataLayer.push({
+          event: "formSubmit",
+          formType: "Teleport Pro Signup",
+        });
+      },
+    };
```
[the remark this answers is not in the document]
C-STYR: changed

utils/cloud-signup.ts
C-STYR: changed

utils/gtm.ts
```diff
@@ -22,8 +22,19 @@ export const GTMEvent = (
       dataLayer.push({
         event,
         ...payload,
+
+        /**eventCallback and eventTimeout are not found in GTM official docs!
+        eventCallback is a function which will execute when all tags which fire on
+        the event have executed; it is scoped to this promise. Always add eventTimeout
+        when you use eventCallback.
+        */
```
iAdramelk, line 30: It was documented in the past. Not sure why is it not documented anymore. :(

components/CloudSignup/CloudSignup.tsx
```diff
@@ -3,13 +3,20 @@ import styled from "styled-components";
...
 
   useEffect(() => {
+    const API_URL = process.env.NEXT_PUBLIC_TELEPORT_API_URL;
+
     if (ref) {
-      renderSignupRequestForm(ref.current);
+      renderSignupRequestForm(ref.current, {
+        apiUrl: API_URL,
+        onSuccess: () =>
+          GTMEvent("formSubmit", { formType: "Teleport Pro Signup" }),
```
iAdramelk, line 18: You maaay want to add braces around the call just to be safe. It will not break anything right now, but GTMEvent return `Promise` (so you can optionally wait until event is sent and callback is fired), and while returning promise here is working now, who knows how GTM code will change in the future.

```suggestion
          { GTMEvent("formSubmit", { formType: "Teleport Pro Signup" }); },
```

iAdramelk (approved):

PR #908, opened 2022-02-15
Adds comments to Marketo-related files

C-STYR:


components/MarketoForm/helpers.ts
```diff
@@ -162,6 +174,10 @@ export const useMarketoForm = (
     [id, setError, thankYou, getToken]
   );
 
+  /**
+   * Runs any time the value of submitted changes after a 5-sec wait period
+   * allowing multiple submissions without page reload.
+   */
```
iAdramelk, line 180: This one is mostly about the `submitted` status for the from. We usually show as a text it near the from of under it.

Like this:

- Submit process ended.
- Set submitted status and write "Submitted!" below the form.
- Wait 5 seconds and remove submitted status together with message.

technically it does not affect form submission in any way except showing the message.

utils/recaptcha.ts
```diff
@@ -122,6 +129,10 @@ export const useRecaptcha = (UID: string) => {
     };
   }, [UID, callbackFNName]);
 
+  /**
+   * This is the memoized return value of useRecaptcha and is destructured
+   * in useMarketoForm
+   */
```
iAdramelk, line 135: Strictly saying, it can be used anywhere, not only in Marketo forms. If we migrate to other from engine, but still want to use recaptcha, this function will stay the same.

iAdramelk (approved):

PR #936, opened 2022-02-28
Update link for security disclosures

wadells:
## Summary

This PR updates our security disclosure link from a defunct zendesk topic to our Teleport Release notes.

Linking our release notes isn't great, as it means there is a lot of sifting involved to find security notes, and it only covers product notes (e.g. no cloud or corp stuff).  However it is vastly better than linking the nearly empty zendesk topic.

A while back, I drafted a longer-term proposal here:  https://gravitational.slab.com/posts/2021-08-security-disclosure-rfd-2fvbunkd However I haven't ever found the priority/will to implement it.  I'm hoping with @reedloden joining, perhaps we'll find that priority.


reedloden (approved):
Seems reasonable to me, and I actually noticed this last week when I was checking out existing security documentation on the website. Thanks for fixing!

C-STYR (approved):
Looks fine from a technical standpoint.  Link works.

PR #947, opened 2022-03-03
Remove docs-related files

iAdramelk:


.eslintrc.json
```diff
@@ -0,0 +1,7 @@
+{
```
C-STYR, line 1: why the change to json format?

components/Dropdown/Dropdown.tsx
```diff
@@ -26,9 +26,7 @@ export type DropdownProps<T> = {
   needBg?: boolean;
 } & BoxProps;
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
C-STYR, line 29: was the `disable` command non-functional as written, or is it no longer needed, or removed for some other reason?

next.config.mjs
```diff
@@ -1,52 +1,53 @@
...
+      permanent: true,
+    },
+    {
+      source: "/login/",
+      destination: "https://teleport.sh",
+      permanent: true,
+    },
+  ],
   rewrites: async () => [
+    {
```
C-STYR, line 25: I have high hopes for this rewrite, but using identical syntax we could not get it to work properly with the blog. Fingers crossed.

next.config.mjs
```diff
@@ -1,52 +1,53 @@
...
+      source: "/docs/:path*/",
+      destination: `${process.env.DOCS_DOMAIN}/docs/:path*/`,
+      basePath: false,
+    },
     {
       source: "/docs/:path*",
```
C-STYR, line 31: does this really require two different rewrites for the trailing slash? 
how about: 
`source: "/docs/:path*/?",` or whatever the proper regex is (going off memory here...)

C-STYR (approved):

.eslintrc.json
iAdramelk: Just a preference. Usually you need `*.js` if you need to write code inside the config and `*.json` if just setting the options is enough.

components/Dropdown/Dropdown.tsx
iAdramelk: I switched to default config that comes with Next.js instead of the custom one. This config does not have this rule set by default.

PR #955, opened 2022-03-07
modern pam tech page

deliaconstantino:
- creates download page for "Modern PAM Buyers Guide" tech paper - preview [here](https://teleport-docs-next-git-delia-modern-pam-tech-page-gravitational.vercel.app/resources/white-papers/pam-buyers-guide/)
- Closes this asana issue: https://app.asana.com/0/1201922005233459/1201922182803846

### `PromoPointsBlock` updates
- refactors `PromoPointsBlock.Item` to reflect actual usage of code:
    - in other tech papers the default fontSize of "16px" was being inherited at all screen sizes, despite the code suggesting that this component should render at `fontSize={["text-lg", "header-3"]}` which translates to `["16px", "24px"]`
    - Updated fontSize to "inherit" and added ability to render a custom fontSize
    - in this use case, rendered at "15px" to avoid a hanging word


components/PromoPointsBlock/PromoPointsBlock.tsx
```diff
@@ -23,13 +25,21 @@ const PromoPointsBlockItem = ({
...
-        {children}
-      </Box>
+      <StyledBox {...props}>{children}</StyledBox>
     </Flex>
   );
 };
 
+const StyledBox = styled(Box)(
+  css({
+    "& p": {
```
sandylcruz, line 35: Should there be a space here?

deliaconstantino: Yes, the styles do not get applied without the space. 

Just took a look though the code base. It looks like targeting html selectors requires the space, but targeting other properties like `hover` do not use a space:
Screen Shot 2022-03-07 at 4 30 50 PM

sandylcruz (approved):

PR #971, opened 2022-03-14
Implement New About Page

sandylcruz:
* Preview [here](https://teleport-docs-next-git-sandy-about-gravitational.vercel.app/about/)
* Uses [`react-grid-system`](https://sealninja.github.io/react-grid-system/)

---
  - https://app.asana.com/0/0/1201921728257445


pages/about/index.tsx
```diff
@@ -0,0 +1,579 @@
...
+}
+
+const leadershipList: Array<LeaderEntry> = [
+  {
+    name: "Ev Kontsevoy",
+    source: EvKontsevoy,
+    title: "Co-founder and CEO",
+  },
+  {
+    name: "Alexander Klizentas",
```
C-STYR, line 436: Please verify spelling of last name

pages/about/index.tsx
```diff
@@ -0,0 +1,579 @@
...
+    title: "VP Sales",
+  },
+  {
+    name: "Phil Simpson",
+    source: PhilSimpson,
+    title: "VP Alliances",
+  },
+  {
+    name: "Reed Lodan",
+    source: ReedLodan,
```
C-STYR, line 472: Please verify spelling of last name

pages/about/index.tsx
sandylcruz: Updated, the design was misspelled.

pages/about/index.tsx
sandylcruz: same here

C-STYR (approved):
This all looks fine, but just a couple of ideas; if you want to implement them, your call: 
1. move the various array constants to their own file and import them
2. instead of a long list of imports, use a variation of `import * as portraits from "./assets/..."`, then call out `{portrait.AnandeliaFadeev}`, etc.

sandylcruz:
@C-STYR I tried extracting the constants to their own data file in the pages folder as well as using the `import * as  x` variation, but kept getting the ["page without valid react component"](https://nextjs.org/docs/messages/page-without-valid-component) build error. The possible solutions to that were to either

1. Move data into `components` folder with the corresponding component
2. Implement [custom page extensions](https://nextjs.org/docs/api-reference/next.config.js/custom-page-extensions) in `next.config` to allow for non-page files to be in the pages folder. This had wider implications with all of the existing `index.mdx` and `index.tsx` files in `pages`, so for now I kept imports as-is.

PR #1037, opened 2022-04-04
Fix signup page title saying `pro` instead of `cloud`

mcbattirola:
This PR changes the title of https://goteleport.com/signup/ from `Signup for Teleport Pro` to `Signup for Teleport Cloud`

As pointed out by @zmb3 on Slack, this is the page to get a cloud account but the title says Teleport Pro


deliaconstantino (approved):

pages/signup/index.tsx
```diff
@@ -10,7 +10,7 @@ import CloudSignup from "components/CloudSignup";
 const Signup = () => {
   return (
     <Pattern1
-      headTitle="Signup for Teleport Pro"
+      headTitle="Signup for Teleport Cloud"
```
zmb3, line 13: I'm pretty sure we want to change to "Sign up" (two words, not one)

PR #1052, opened 2022-04-07
Fix button flash on page load

deliaconstantino:
- move button transition to pseudo elements


C-STYR (commented):
OK, I've had a little time to look into this.  First off, well done for spotting the issue.  The flash is so brief on my machine I hadn't noticed it.  
Let's rework this solution.  Your fix, which is not to display the button until destyling is complete, works but has a couple of problems:
1. Toggling the `display` property has implications for CLS - which aren't hugely significant in this instance but as a general point of principle we should be toggling `visibility` instead
2. We are getting an extra render (this isn't the result of your code, it existed previously, but we should fix it if we can now). The extra render happens because we are changing the value of the `loaded` prop which we pass into `StyledFormWrapper`. 
Have another crack at this and see if we can simplify.

deliaconstantino:
@C-STYR 
1. I looked further into this based on your comments. I was not able to get `visibility="hidden"` to have the desired effect - the green flash still shows, though a bit more briefly than it currently shows on the live site without `visibility="hidden"`. The green flash is not visible with `display="none"`. I took some videos and slowed them down to clarify what is happening on page load:

**Display="none"** - no green flash:

https://user-images.githubusercontent.com/70108137/163585086-16a284f8-66ae-4dc2-a561-42d8d978b423.mp4

**Visibility="hidden"** - green flash:

https://user-images.githubusercontent.com/70108137/163585125-0d3db31f-a52a-4efa-9c2e-b99f942cdd74.mp4

2. I was not able to find a successful solution that removes a re-render. I also looked at your solution on [cole/TEST](https://github.com/gravitational/next/pull/1056)  and noticed that the green flash is still visible there - added a comment. One interesting note is that my solution added a fourth render, as captured by the console log of "loaded" in `<Marketo />`, while your solution in cole/TEST and the live code has 3 renders. I believe the 4th render could be related to the additional state value (`destyled`) being set here, though I am unsure how that would trigger a re-render:
Screen Shot 2022-04-15 at 10 12 49 AM

Screen Shot 2022-04-15 at 10 12 11 AM

PR #1081, opened 2022-04-20
Center component on features/access-requests

deliaconstantino:


sandylcruz:
This PR is good, though I did notice that the UI is weird in that there's differing choices made about what to do with image placement on mobile screens:

* In `SectionVisuals`, the image goes below the title/description
* In `SectionVideo`, the video goes above the title/description

Screen Shot 2022-04-20 at 2 29 15 PM

This is outside the scope of your PR, but a consensus should probably be made so everything is consistent.

sandylcruz (approved):

deliaconstantino:
> This PR is good, though I did notice that the UI is weird in that there's differing choices made about what to do with image placement on mobile screens:
> 
> * In `SectionVisuals`, the image goes below the title/description
> * In `SectionVideo`, the video goes above the title/description
> 
> This is outside the scope of your PR, but a consensus should probably be made so everything is consistent.

Agreed, this does look a bit strange with these two components paired together. I'll open a ticket in Asana so it can be discussed.

PR #1101, opened 2022-04-28
New KubeCon Happy Hour Page

C-STYR:
Updates `/kubecon-happy-hour` after redesign request
Preview [here](https://teleport-docs-next-git-cole-kube-gravitational.vercel.app/kubecon-happy-hour)

---
  - https://app.asana.com/0/0/1202191669556301


jeffgaynor (approved):
If a smaller image is possible, it would be good to help reduce repo bloat. Even at this size, it can be compressed (lossless) a bit:
image

Otherwise, looks fine from a code perspective.

C-STYR:
> If a smaller image is possible, it would be good to help reduce repo bloat. Even at this size, it can be compressed (lossless) a bit: image
> 
> Otherwise, looks fine from a code perspective.

Updated - this page will only run for 3 weeks, but I did cut down the shipped-to-the-browser version to 150kb.

PR #1105, opened 2022-04-29
marketo flash fadein

deliaconstantino:


C-STYR:
Screen Shot 2022-05-04 at 5 09 39 AM
With the sped up fade-in I'm still seeing the green flash on Security Visionaries and some other pages.  Let's restore the original speed.  Also, see screenshot above: the form is occluding the down caret at this screen width.

PR #1147, opened 2022-05-09
Ports `nginx` redirects to Next - final version

C-STYR:
- adds `/redirects` directory in `/server`
- ports `nginx-redirects.conf` rules from `/web` AND revises rules and fixes redirect loops & contradictions
`docs` and `blog` rules will be added to those repos separately
- adds header rules from `nginx.conf`


server/redirects/legacy-redirects.mjs
```diff
@@ -0,0 +1,640 @@
...
+      {
+        source: "/tags",
+        destination: "/blog/",
+        permanent: true
+      },
+
+      //DOCS REDIRECTS
+      {
+        source: "/teleport/docs/architecture",
+        destination: "https://goteleport.com/docs/architecture/overview/",
```
reedloden, line 146: Shouldn't these (and the ones below) be relative links? Having them as absolute means we can't do proper testing on a separate staging site.

server/redirects/legacy-redirects.mjs
```diff
@@ -0,0 +1,640 @@
...
+        destination: "https://goteleport.com/docs/architecture/authentication/",
+        permanent: true
+      },
+      {
+        source: "/teleport/docs/architecture/teleport_auth",
+        destination: "https://goteleport.com/docs/architecture/authentication/",
+        permanent: true
+      },
+      {
+        source: "/enterprise/introduction.md",
```
reedloden, line 185: .md seems odd? Should this have been just `/enterprise/introduction`?

server/redirects/legacy-redirects.mjs
```diff
@@ -0,0 +1,640 @@
...
+        permanent: true,
+      },
+      {
+        source: "/resources/videos/gravity-6-ui-tour",
+        destination: "/resources/videos/",
+        permanent: true,
+      },
+      {
+        source: "/resources/articles",
+        destination: "https://goteleport.com/blog/",
```
reedloden, line 326: /blog/

server/redirects/legacy-redirects.mjs
```diff
@@ -0,0 +1,640 @@
...
+        permanent: true,
+      },
+      {
+        source: "/resources/webcasts(/.*?)", 
+        destination: "/resources/",
+        permanent: true,
+      },
+      {
+        source: "/resources/videos/application-access/docs/enterprise/ssh-kubernetes-fedramp", 
+        destination: "https://goteleport.com/docs/enterprise/fedramp/",
```
reedloden, line 341: /docs/enterprise/fedramp/

server/redirects/legacy-redirects.mjs
```diff
@@ -0,0 +1,640 @@
...
+      
+      
+      //HOW IT WORKS REDIRECTS
+      {
+        source: "/teleport/how-it-works",
+        destination: "/how-it-works/",
+        permanent: true
+      },
+      {
+        source: "/teleport/how-it-works/shh-kubernetes-access-fedramp-soc2",
```
reedloden, line 535: Seems like a typo... should be ssh instead of shh

jeffgaynor (approved):
Looks good from here.

server/redirects/legacy-redirects.mjs
C-STYR: these (and the ones that follow) are cross-repo redirects so must be either absolute or dynamic (unless I misunderstand how Cloudflare works); I considered making them dynamic but we because we don't have a staging site presently I would have had to update them later anyway; let's talk about this in the meeting.

reedloden (approved):

PR #1152, opened 2022-05-10
Add gluu

deliaconstantino:
- preview [here](https://next-git-delia-gluu-gravitational.vercel.app/case-study/)
Screen Shot 2022-05-11 at 9 57 13 AM
---

  - https://app.asana.com/0/0/1202248282468156


sandylcruz:
Should the industry be added for Gluu? I can't find where we're using the industry information anywhere, though I'm not sure if it'll be needed in the future.

sandylcruz (approved):

deliaconstantino:
> Should the industry be added for Gluu? I can't find where we're using the industry information anywhere, though I'm not sure if it'll be needed in the future.

Industry is currently used for the "filter by industry" tab. Currently there's one other without an industry, but I can verify if it should be added for gluu.
Screen Shot 2022-05-11 at 4 12 34 PM

PR #1180, opened 2022-05-16
Address console warning on use-cases/aws 

deliaconstantino:
# Problem:
- warning in console about `sizes` prop
image

# Solution
- removes unneeded `sizes` prop from `Image` component using `layout="intrinsic"`
- Per next/image [sizes documentation](https://nextjs.org/docs/api-reference/next/image#sizes):
> If you are using layout="intrinsic" or layout="fixed", then sizes is not needed because the upper bound width is constrained already.

# Accessibility and Performance improvements for `/use-cases/aws` page:
- adds descriptive `alt` tags for images
- adds `alt=""` for [decorative images](https://www.w3.org/WAI/tutorials/images/decorative/)
- converts hero imago to a direct import in order to add `priority="true"` prop for performance implications 
Screen Shot 2022-05-16 at 11 38 55 AM

- optimizes images
- preview [/use-cases/aws](https://next-git-delia-image-bug-gravitational.vercel.app/use-cases/aws/)

---
  - https://app.asana.com/0/0/1202255011783012


sandylcruz:
Looks good, nice work on researching if the icons needed alts or not.

sandylcruz (approved):

PR #1181, opened 2022-05-16
Creates A/B Sandbox environment

C-STYR:
- disallows indexing past `/sandbox/`
- adds homepage clone at `/sandbox/homepage/`
- adds happy hour clone at `/sandbox/happy-hour/`


public/robots.txt
```diff
@@ -11,4 +11,5 @@ Disallow: /teleport/docs/2.4/
 Disallow: /categories/
 Disallow: /_shared/
 Disallow: /docs/ver/
+Disallow: /sandbox/
```
jeffgaynor, line 14: Does this prevent anything at this path from getting added via our sitemap generator?

C-STYR: no, it just prevents SEO crawlers from indexing it

jeffgaynor (approved):

PR #1182, opened 2022-05-16
Sandbox-related updates

C-STYR:
- prevents sitemap location creation for `/sandbox/*` pages
- adds `nofollow` attributes to `sandbox` test pages


components/Pattern1/Pattern1.tsx
```diff
@@ -23,6 +23,7 @@ export interface Pattern1Props {
   children: React.ReactNode;
   secondLogo?: string;
   previewImage?: string;
+  nofollow?: boolean;
```
sandylcruz, line 26: Should `nofollow` here also be camel case to be consistent with the casing of `noFollow` in `Head`?

sandylcruz (approved):

components/Pattern1/Pattern1.tsx
C-STYR: I'd actually prefer it the other way around, as the actual content type is `nofollow` not `noFollow` but didn't want to change `noIndex` as written so settled on this compromise.

PR #1192, opened 2022-05-20
Update datasheet

deliaconstantino:
- update Teleport-Kubernetes-Datasheet to new url and implement redirect
- preview PDF [here](https://next-git-delia-add-datasheet-gravitational.vercel.app/Teleport-Kubernetes-Datasheet.pdf)
- Note: ran Prettier on `server/redirects/legacy-redirects.mjs` which corrected spacing for the entire file. I verified that everything looked correct, and added a comment below noting the updated code in this PR.
---
  - https://app.asana.com/0/0/1202301262550345


server/redirects/legacy-redirects.mjs
```diff
@@ -14,620 +14,633 @@ export const legacyRedirects = [
...
+  {
+    source: "/pdf/state-of-infrastructure-access-2021.pdf",
+    destination: "/state-of-infrastructure-access-2021.pdf",
+    permanent: true,
+  },
+  {
+    source: "/Teleport-Kubernetes-Datasheet-2022.pdf",
+    destination: "/Teleport-Kubernetes-Datasheet.pdf",
+    permanent: true,
+  },
```
deliaconstantino, line 361: Updated this section:
1. created `// PDF REDIRECTS` section, moving lines 352-356 inside
2. add lines 357-361

jeffgaynor: Since you are creating a redirect, I think it would be better to move those few PDFs into the `/static/resources` folder in an appropriate subfolder and update the redirects accordingly. @C-STYR Any reason that you know of we should not do that?

server/redirects/legacy-redirects.mjs
```diff
@@ -355,7 +355,7 @@ export const legacyRedirects = [
 
   // DATASHEET REDIRECTS
   {
-    source: "/Teleport-Kubernetes-Datasheet-2022.pdf",
+    source: "/Teleport-Kubernetes-Datasheet(-2022)?.pdf",
     destination:
       "/static/resources/datasheets/Teleport-Kubernetes-Datasheet.pdf",
     permanent: true,
```
jeffgaynor, line 361: If you want to be future proof for any year this millennium, try:
```
source: "/Teleport-Kubernetes-Datasheet(-2[0-9]{3})?.pdf"
```

jeffgaynor (approved):
Thanks for making the adjustment. I left a little suggestion, but otherwise, looks good.

server/redirects/legacy-redirects.mjs
deliaconstantino: Thanks for suggesting this, but I'm not sure if it's needed at this point. There was a previous request to add this same PDF when it was appended with 2022. Today's request is updating the filename to remove 2022. I think this is more of a question of clarifying the desired url for any new pages.

